PR #3, opened 2021-08-30
reduce length of default string used in createing host name

mmulholla:
Adding fix already added to charts repo: https://github.com/openshift-helm-charts/charts/pull/314

Reduce the length of the names used for OCP testing.


dperaza4dustbit (approved):
lgtm

PR #12, opened 2021-09-11
Fixing a security issue when running pytest

dperaza4dustbit:
Setting the traceback flag to false to avoid traceback printing
partial github token characters. We will need to rotate all our
github tokens after this PR merges.


dperaza4dustbit:
This is the test that shows traceback is not showed anymore: https://github.com/dperaza4dustbit/development/pull/12/checks. However, are we really OK hiding the complete traceback. Would that help us in a debugging situation?

zonggen (approved):
LGTM :)

PR #16, opened 2021-09-15
test_submitted_charts: add software name to branch names

zonggen:
This avoids collision between test runs of oc change and chart-verifier
change.

### DIFF
--- a/tests/functional/test_submitted_charts.py
+++ b/tests/functional/test_submitted_charts.py
@@ -222,8 +222,9 @@ def submission_tests_run_for_submitted_charts(secrets):
 
         for vendor_type, vendor_name, chart_name, chart_version in secrets.submitted_charts:
             chart_dir = f'charts/{vendor_type}/{vendor_name}/{chart_name}'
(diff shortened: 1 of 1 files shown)


tests/functional/test_submitted_charts.py
```diff
@@ -224,6 +224,11 @@ def submission_tests_run_for_submitted_charts(secrets):
             chart_dir = f'charts/{vendor_type}/{vendor_name}/{chart_name}'
             base_branch = f'{secrets.pr_base_branch}-{vendor_type}-{vendor_name}-{chart_name}-{chart_version}'
             pr_branch = f'{secrets.pr_base_branch}-{vendor_type}-{vendor_name}-{chart_name}-{chart_version}-pr'
+
+            if not secrets.software_name:
```
mmulholla, line 228: Confused you use secrets.software_name is it is not defined?

dperaza4dustbit: This will check is empty or None. Either case yeah what is the point on using below

zonggen: Actually this is redundant, somehow I was thinking during manual trigger the `software-name` is optional. Updating

mmulholla (approved):
LGTM

PR #20, opened 2021-09-22
Add unique vendor name to each test case

dperaza4dustbit:
In order to avoid release name to collide when two PRs are send
for testing at the same time. Vendor it is used as organization
in the release name template.


mmulholla (approved):
LGTM

PR #24, opened 2021-09-23
Fixing workflow release cleanup issue running on forks

dperaza4dustbit:
Expected tag does not match actual tags used for releases

### DIFF
--- a/tests/functional/test_chart_src_with_report.py
+++ b/tests/functional/test_chart_src_with_report.py
@@ -321,8 +321,9 @@ def the_index_yaml_is_updated_with_a_new_entry(secrets):
 def the_release_is_published(secrets):
     """a release is published with the chart"""
 
-    expected_tag = f'{secrets.vendor}-{secrets.chart_name}-{secrets.chart_version}-test-pr{secrets.pr_number}'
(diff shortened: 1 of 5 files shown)


zonggen:
I just realized that that previous change on the release naming might have broken the current code. If you search `test-pr` through the code base, you would find that when the CI workflow is triggered by the bot account, the vendor name has already set to a unique id (PR number in the sandbox, which is a counter and cannot be repeated). This is done not only in the `tests/` code part but also the `scripts/` part. The real issue is to find the root cause for previous repeating release naming bug.

Note: when you see a `test` label on the PR, it will set the environment variable `TRIGGERED_BY_TEST`, and add `-test-{pr_number}` as its suffix in the release.

zonggen:
If you could either add timestamps in other places used `test-pr` (again https://github.com/openshift-helm-charts/charts/pull/294/files) then this disparity of naming should be fixed, but that didn't answer why before the recent fix the release is not cleaned up. Or you could revert the previous fix https://github.com/openshift-helm-charts/development/commit/54ba09ae7b8fc4d032e582990e9e9bcf15a03b56 and we can investigate on why the release was not cleaned up.

dperaza4dustbit:
This is not needed anymore as PR 225 fixes the problem

PR #55, opened 2021-10-14
chartrepomanager: add check for 'charts.openshift.io/name'

zonggen:
Fixes: https://issues.redhat.com/browse/HELM-226

### DIFF
--- a/scripts/src/chartrepomanager/chartrepomanager.py
+++ b/scripts/src/chartrepomanager/chartrepomanager.py
@@ -274,6 +274,13 @@ def update_chart_annotation(category, organization, chart_file_name, chart, repo
 
     annotations = report_info.get_report_annotations(report_path)
 
+    print("[INFO] Check 'charts.openshift.io/name'")
+    # Fixes https://issues.redhat.com/browse/HELM-226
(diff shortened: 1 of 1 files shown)


scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -274,6 +274,13 @@ def update_chart_annotation(category, organization, chart_file_name, chart, repo
 
     annotations = report_info.get_report_annotations(report_path)
 
+    print("[INFO] Check 'charts.openshift.io/name'")
+    # Fixes https://issues.redhat.com/browse/HELM-226
+    if 'charts.openshift.io/name' not in annotations:
+        print("[ERROR] 'charts.openshift.io/name' is required in chart annotations.")
+        print("[ERROR] See https://helm.sh/docs/topics/charts/#the-chartyaml-file")
+        sys.exit(1)
+
```
dperaza4dustbit, line 283: Lets discuss this, want to understand why it should not be a check in chart verifier that can be run outside the context of the certification program.

zonggen:
As per discussion in the scrum meeting, we indeed want this check to be in `chart-verifier` and use profile support to enforce different checks.

PR #57, opened 2021-10-20
prevent cron job running on forks

mmulholla:
Made it so the cron job will only run from the openshift-helm-charts/charts repository.

workflow dispatch is not restricted.

### DIFF
--- a/.github/workflows/schedule.yml
+++ b/.github/workflows/schedule.yml
@@ -40,20 +40,20 @@ jobs:
           fi
 
       - name: Install oc
-        if: steps.check_main.outputs.run-job == 'true'
+        if: steps.check_repo.outputs.run-job == 'true'
         run: |
           curl -sLO https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule.yml
```diff
@@ -25,18 +25,35 @@ jobs:
...
+          if [ $GITHUB_EVENT_NAME == 'workflow_dispatch' ]; then
+            echo '::set-output name=run-job::true'
+          elif [ $GITHUB_REPOSITORY == "openshift-helm-charts/charts" ]; then
+            echo '::set-output name=run-job::true'
+          else
+            echo '::set-output name=run-job::false'
+          fi
+
       - name: Install oc
+        if: steps.check_main.outputs.run-job == 'true'
```
zonggen, line 43: Think this should be `check_repo`, and the following `check_main`s.

zonggen:
Can you rebase on top of latest 'main' branch?

zonggen (approved):
Tests passed and code LGTM!

dperaza4dustbit (approved):
lgtm

PR #58, opened 2021-10-26
tests: refactor tests to use classes and split utils.py into multiple files

zonggen:
Use python classes to increase the code reusability and extendability.

Closes:https://issues.redhat.com/browse/HELM-246


tests/functional/step_defs/test_chart_src_with_report.py
```diff
@@ -0,0 +1,85 @@
...
+    given,
+    scenario,
+    then,
+    when,
+)
+
+from functional.utils.chart_certification import ChartCertificationE2ETestSingle
+
+@pytest.fixture
+def workflow_test():
```
mmulholla, line 18: Would like to see these values passed in as parameters from the bdd file. This would enable us to add new tests without modifying or duplicating python files.

tests/functional/step_defs/test_chart_src_with_report.py
```diff
@@ -0,0 +1,85 @@
...
+def test_partner_chart_src_submission():
+    """The partner hashicorp submits an error-free chart source with report for vault."""
+
+
+@scenario('../features/report_and_chart_src.feature', "A redhat associate submits an error-free chart source with report for vault")
+def test_redhat_chart_src_submission():
+    """A redhat associate submits an error-free chart source with report for vault."""
+
+
+@given("hashicorp is a valid partner")
```
mmulholla, line 37: For example, this ```@given``` function is repeated in multiple .py files. We should make these common too, by including the tests in the same bdd file.

mmulholla (changes requested):
I really like the refactoring so far, very good work! To complete I think we need to refactor the bdd files and remove even more test specific code from the tests.

tests/functional/step_defs/test_chart_src_with_report.py
zonggen: Just to make sure I have the correct understanding:

Refactoring the bdd files means to merge the current bdd files together and use a table for parameter settings, for example: 

  | vendor_type  | vendor    | chart   | version  |
  | partners     | hashicorp | vault   | 0.13.0   |
  | redhat       | redhat    | vault   | 0.13.0   |

and other parameters such as with/without tarball/source, with/without report?

zonggen: Or do we want just a table for the parameters that are set in the following code:
```
    test_name = 'Test Chart Source With Report'
    test_chart = 'tests/data/vault-0.13.0.tgz'
    test_report = 'tests/data/report.yaml'
```

Essentially I am asking regarding the degree of refactoring for bdd files. I tend to propose we make the current bdd files separate but parameterize the `@given` and other constant to a table in bdd files.

mmulholla: Yes, refactoring the bdd files to add a parameter settings.

with/without tarball/source  - don't think these need parameters so just different givens.

For the bdd files, it would make sense to review current content, maybe some can be combined. I am not averse to one big bdd file if is makes most sense

Also, to be honest with bdd files there are a ton of different ways to do them and I am no authority but the more re-use and flexibilty we get the better. As I mention, the goal here is that we can add tests just by changing the bdd file.  
- New chart : just add it as params to the bbd file.
- New scenario : just create a flow hopefully using existing existing given, when, then but in a different combination. Some tests might need to add one or two but mostly re-use.

zonggen: Got it, thanks for explaining!

mmulholla (approved):
LGTM. Good work!

tests/functional/utils/chart_certification.py
```diff
@@ -0,0 +1,752 @@
...
+
+        r = github_api(
+            'get', f'repos/{self.secrets.test_repo}/pulls/{pr_number}/merge', self.secrets.bot_token)
+        if r.status_code == 204:
+            logging.info(f"PR for {chart} merged sucessfully")
+        else:
+            logging.warning(
+                f"Workflow for submitted PR success but PR not merged, chart: {chart}")
+            return
+
```
tisutisu, line 632: can we reuse the check_pull_request_result() function as defined in "ChartCertificationE2ETestSingle" data class and make check_pull_request_result() as part of "ChartCertificationE2ETest" data class ? Above portion of code looks similar.

tisutisu (changes requested):
Overall the refactoring looks good, I have added few comments, please have a look.

tests/functional/utils/chart_certification.py
```diff
@@ -0,0 +1,752 @@
...
+        logging.info(f"Delete '{self.secrets.test_repo}:{self.secrets.pr_branch}'")
+        github_api(
+            'delete', f'repos/{self.secrets.test_repo}/git/refs/heads/{self.secrets.pr_branch}', self.secrets.bot_token)
+
+        logging.info(f"Delete local '{self.secrets.base_branch}'")
+        try:
+            self.repo.git.branch('-D', self.secrets.base_branch)
+        except git.exc.GitCommandError:
+            logging.info(f"Local '{self.secrets.base_branch}' does not exist")
+
```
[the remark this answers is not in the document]
zonggen: Updated, the rewritten cleanup() should take care of https://issues.redhat.com/browse/HELM-256.

tests/functional/utils/chart_certification.py
```diff
@@ -0,0 +1,752 @@
...
+        r = github_api(
+            'get', f'repos/{self.secrets.test_repo}/pulls/{pr_number}/merge', self.secrets.bot_token)
+        if r.status_code == 204:
+            logging.info(f"PR for {chart} merged sucessfully")
+        else:
+            logging.warning(
+                f"Workflow for submitted PR success but PR not merged, chart: {chart}")
+            return
+
+        # Check index.yaml is updated
```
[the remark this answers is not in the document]
zonggen: Updated

tests/functional/utils/chart_certification.py
```diff
@@ -0,0 +1,752 @@
...
+            logging.warning(
+                f"{chart} not added to index")
+            return
+
+        logging.info(
+            f"Index updated correctly for {chart}, cleaning up local branch")
+        repo.git.checkout(old_branch)
+        repo.git.branch('-D', f'{base_branch}-gh-pages')
+
+        # Check release is published
```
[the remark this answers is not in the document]
zonggen: Updated

tests/functional/utils/chart_certification.py
zonggen: Lifted the common methods into the parent class. Updated

PR #63, opened 2021-11-08
Fix problem with involing report in cert flow

mmulholla:
Fix problem invoking chart verifier report command fron cert flow


dperaza4dustbit (approved):
lgtm, not merging cause of test failures

dperaza4dustbit:
Oh, taking a closer look, it looks like all test passed, merging

PR #64, opened 2021-11-09
Diff oc login if not main repo

mmulholla:
specify --insecure-skip-tls-verify  on oc login if not main repo.

This should help us testing on fork, let me know if you think it should be more restrictive.


dperaza4dustbit (approved):
lgtm

PR #65, opened 2021-11-10
Added tests for community path 

tisutisu:
This PR will cover story : https://issues.redhat.com/browse/HELM-245


zonggen (commented):
LGTM, though we really want to parallelize the tests from now on since one single test workflow run takes ~35min to finish. The number of tests are increased quadratically O(nm), where n is the number of parameters we set in feature file and m is the number of scenarios. @dperaza4dustbit

.github/workflows/test.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_BODY: "Test triggered by ${{ github.event.pull_request.html_url }}."
         run: |
-          ve1/bin/pytest tests/ --log-cli-level=WARNING --ignore=tests/functional/step_defs/test_submitted_charts.py --tb=short
+          ve1/bin/pytest tests/ -n 2 --log-cli-level=WARNING --ignore=tests/functional/step_defs/test_submitted_charts.py --tb=short
```
zonggen, line 85: Thanks for making this change, but I think this won't work because the collision on release tags and names. Currently we have `expected_tag = f'{vendor}-{chart_name}-{chart_version}'` and `expected_chart_asset = f'{vendor}-{chart_tgz}'`, and vendor is differentiated by PR number. Right now the tests won't collide between PR tests but _will_ collide within test cases of the same PR.

In order to mitigate this collision we need a unique release name and release tag even between test cases of the same PR. I think this can be a separate story after the community path is merged. WDYT @tisutisu @dperaza4dustbit  @mmulholla ?

tisutisu: I think we can make the changes related to parallel running of the tests in separate PR.

mmulholla: Looks like David addressed this: https://github.com/openshift-helm-charts/development/pull/66

tests/functional/step_defs/test_chart_src_with_report.py
```diff
@@ -73,3 +76,14 @@ def index_yaml_is_updated_with_new_entry(workflow_test):
 def release_is_published(workflow_test):
     """a release is published with corresponding report and chart tarball."""
     workflow_test.check_release_result()
+
+@then("the pull request is not merged")
```
mmulholla, line 80: These methods appear in several files which I would like to get away from. If it means we have one big feature file I do not object to that although maybe one python routine could import several features.

mmulholla (commented):
LGTM apart from the repeated functions, would like to see if we can reduce that.

zonggen (approved):
LGTM!

tests/functional/step_defs/test_chart_src_with_report.py
tisutisu: Merging this PR, we can discuss these improvements and initiate another PR.

PR #69, opened 2021-11-17
[Old one] Added chart verifier comes back with failure test scenario

tisutisu:
This PR will cover story : https://issues.redhat.com/browse/HELM-243


zonggen (commented):
From HELM-243 it seems like for `redhat` workflow we need to check the `providerType=community` as a result (see an example here: https://github.com/openshift-helm-charts/charts/blob/gh-pages/index.yaml#L961). So in addition to checking there's an entry, we want to ensure the entry is added as a `community` type in index.yaml.

Otherwise looks good to me.

zonggen:
Code LGTM, there was one failure in the test run. Re-triggered test workflow to see if it passes.

PR #72, opened 2021-11-19
Fix community and redhat path when kubeVersion not set in chart

dperaza4dustbit:
We are not checking if kubeVersion is in chart before trying to get the value. This is causing redhat and comunity charts that dont have kubeVersion set to fail on release step.


zonggen (approved):
LGTM.

Test failed with `/home/runner/work/_temp/3767ae8a-7c86-4cd9-af4f-0c1815ee45c3.sh: line 9: unexpected EOF while looking for matching `''` , but we don't have shell script in the test suite and I don't think this PR induces the failure.

PR #73, opened 2021-11-19
.github: fix unmatched quotes in workflows

zonggen:
Caused by unpredicted single(')/double(") quotes in PR body, shell strings in
workflow produce unmatched quotes error. This change ensures we parse
the string correctly before passing as arguments.

Refer: https://github.com/openshift-helm-charts/development/runs/4266289403?check_suite_focus=true#step:7:14


dperaza4dustbit (approved):
lgtm

PR #74, opened 2021-11-22
Added sha value does not match scenario

tisutisu:
This PR will cover story: https://issues.redhat.com/browse/HELM-265

### DIFF
--- a/tests/functional/features/chart_verifier_comes_back_with_failures.feature
+++ b/tests/functional/features/chart_verifier_comes_back_with_failures.feature
@@ -1,0 +1,43 @@
+Feature: Chart verifier comes back with a failure
+  Partners, redhat or community user submit charts which does not contain README file
+
+  Examples:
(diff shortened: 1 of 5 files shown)


zonggen:
Code LGTM, re-triggering test to see if the test case still fails.

zonggen:
Just want to confirm, did you piggy back the test chart version change to this PR? Can you rebase on top of main so the changes HELM265 only? I this is ready to go after rebasing, just need attention from @dperaza4dustbit  and @mmulholla

tisutisu:
> Just want to confirm, did you piggy back the test chart version change to this PR? Can you rebase on top of main so the changes are HELM265 only? I think this is ready to go after rebasing, just need attention from @dperaza4dustbit and @mmulholla

Looks like I messed up this. Will close this and create a new PR with only for the HELM-265 scenario.

PR #77, opened 2021-11-26
workflow: parallelize pytest in the github runner

zonggen:
Currently the a single runthrough of our e2e test takes ~35mins and is growing quadratically. This change uses pytest-xdist plugin to parallelize the pytest runs. Computational limitations apply, as Github linux runners only have 2 cores CPU:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

Note: Since the `-n auto` option assigns each core one test to run, this option limits to 2 threads (2 cores).

EDIT:

1. 
```
parallel testing:
 - n=1: ~34m58s
 - n=2: ~17m45s
 - n=4: ~9m33s
 - n=8 failed (exceeded api call limit): {'message': 'API rate limit exceeded for user ID 83200018.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
```
Therefore it is reasonable to set n=4. See a test run here: https://github.com/zonggen/development/runs/4357928277?check_suite_focus=true

2. Change timestamp to uuid, since for parallel runs the timestamp might be identical, even with time_ns().

Closes: https://issues.redhat.com/browse/HELM-267


tests/functional/utils/chart_certification.py
```diff
@@ -348,8 +357,11 @@ def update_test_report(self, test_report):
             self.__post_init__()
 
     def get_unique_vendor(self, vendor):
+        """Set unique vendor name.
+        Note that release tag is generated with this vendor name.
+        """
         # unique string based on current time in seconds
-        suffix = str(int(time.time()))
```
dperaza4dustbit, line 352: what is wrong with time guid is big string, are we going to have collisions within milliseconds. Update the comment if you go with guid

tests/functional/utils/chart_certification.py
```diff
@@ -289,7 +292,7 @@ def __post_init__(self) -> None:
                     f'HEAD:refs/heads/{current_branch}', '-f')
 
         pretty_test_name = self.test_name.strip().lower().replace(' ', '-')
-        base_branch = f'{pretty_test_name}-{current_branch}' if pretty_test_name else f'test-{current_branch}'
+        base_branch = f'{self.uuid}-{pretty_test_name}-{current_branch}' if pretty_test_name else f'{self.uuid}-test-{current_branch}'
```
dperaza4dustbit, line 295: uuids are very big strings is it necessary for uniqueness?

.github/workflows/schedule.yml
```diff
@@ -324,4 +324,4 @@ jobs:
           printf "[INFO] Notify ID: '%s'\n" "${{ env.NOTIFY_ID }}"
           printf "[INFO] Software Name: '%s'\n" "${{ env.SOFTWARE_NAME }}"
           printf "[INFO] Software Version: '%s'\n" "${{ env.SOFTWARE_VERSION }}"
-          ve1/bin/pytest tests/functional/step_defs/test_submitted_charts.py --log-cli-level=WARNING --tb=short
+          ve1/bin/pytest -n 4 tests/functional/step_defs/test_submitted_charts.py --log-cli-level=WARNING --tb=short
```
dperaza4dustbit, line 327: where does 4 comes from why not 8 or 16 etc...

dperaza4dustbit (commented):
looks good, just few questions

tests/functional/utils/chart_certification.py
zonggen: For parallel runs the timestamp might be identical, even with time_ns() (nanoseconds). The first batch of tests (the first job from each thread) have a higher chance of collision, I noticed the collision even using the nanosecond timestamp, and after switching to uuid I don't see collisions. Therefore I used uuid, great question.

updating comments to explain this change.

tests/functional/utils/chart_certification.py
zonggen: see previous comment :)

.github/workflows/schedule.yml
zonggen:
```
parallel testing:
 - n=1: ~34m58s
 - n=2: ~17m45s
 - n=4: ~9m33s
 - n=8 failed (exceeded api call limit): {'message': 'API rate limit exceeded for user ID 83200018.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
```
When n>=8 we are easier to hit API rate limit from Github (something we can optimize later to minimize API call footprint). Therefore it is reasonable to set n=4. See a test run here: https://github.com/zonggen/development/runs/4357928277?check_suite_focus=true

zonggen: again, great question. thanks for reviewing!

mmulholla, line 327: Would be good to set 4 as an environment variable so it is set in just one place. Then if we ever want to try different numbers in needs just a single edit. Guess a counter to that would be if see a future need to use different number for different pytest invocations.

mmulholla (commented):
Maybe a question more than a comment, but a several dependencies were  added to requirements.txt. Are most need under the covers by pytest to enable multiple runners?

.github/workflows/schedule.yml
zonggen: Agreed. I realized we don't need to support parallelization for submitted chart testing because there's only one pytest file with one test case. After removing, there was only one ocurrence of this value n=4. Though I strongly agree we should deduplicate as much as possible.

zonggen:
> Are most need under the covers by pytest to enable multiple runners?

Was curious too so I started a new empty python virutal environment, pip install the requirement.txt and pip install the pytest-xdist module then dump the dependencies into the requirement.txt. So I assume those are requied by/dependency of the pytest-xdist package.

tests/functional/utils/chart_certification.py
dperaza4dustbit: OK, thanks for explanation

dperaza4dustbit (approved):
lgtm

mmulholla (approved):
LGTM

PR #78, opened 2021-11-29
Added chart-verifier comes back with failure scenario

tisutisu:
This PR will cover story : https://issues.redhat.com/browse/HELM-243

### DIFF
--- a/tests/functional/utils/chart_certification.py
+++ b/tests/functional/utils/chart_certification.py
@@ -426,6 +426,14 @@ class ChartCertificationE2ETestSingle(ChartCertificationE2ETest):
                 except Exception as e:
                     pytest.fail("Failed to update version in yaml file")
 
+    def remove_readme_file(self):
+        with SetDirectory(Path(self.temp_dir.name)):
+            path = f'{self.chart_directory}/{self.secrets.chart_version}/src/README.md'
+            try:
+                os.remove(path)
+            except Exception as e:
+                pytest.fail(f"Failed to remove readme file : {e}")
+
     def process_owners_file(self):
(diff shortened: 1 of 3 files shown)


tests/functional/utils/chart_certification.py
```diff
@@ -171,6 +171,12 @@ def check_index_yaml(self, base_branch, vendor, chart_name, chart_version, logge
                 f"{chart_name} {chart_version} not added to index")
             return False
 
+        if check_provider_type and self.secrets.vendor_type == 'redhat':
```
zonggen, line 174: Can you add a comment on why we are checking this condition? A comment helps when anyone later looks at this condition and wondering why we did this. Also this condition might or might not change so it's beneficial to have a record.

zonggen:
Minor nit otherwise LGTM.

tests/functional/utils/chart_certification.py
tisutisu: Agreed. Added comments explaining this check.

zonggen: Thanks! resolving.

zonggen:
Comment was addressed, LGTM.

zonggen (approved):

mmulholla (commented):
The first and 3rd scenario are identical apart from their inputs. Consider reducing to 1.

tisutisu:
> The first and 3rd scenario are identical apart from their inputs. Consider reducing to 1.

Thanks for pointing this out. Merged the first and 3rd scenario into one.

dperaza4dustbit (approved):
LGTM

mmulholla (approved):

PR #79, opened 2021-11-29
Adding script to create admin sa with permanent token duration

dperaza4dustbit:
Tested token produced by this script here: https://github.com/dperaza4dustbit/development/runs/4360138781?check_suite_focus=true

Signed-off-by: David Peraza <dperaza@redhat.com>

### DIFF
--- a/scripts/src/saforcertadmin/cluster_role_binding.yaml
+++ b/scripts/src/saforcertadmin/cluster_role_binding.yaml
@@ -1,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: rh-cert-user-cluster-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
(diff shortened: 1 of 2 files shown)


mmulholla:
Could we run this as a schedule activity to reset the tokens every so often? Or maybe create a 3 month token and schedule this to run every 90 days?

mmulholla:
Here is a github action which enable the setting/updating of repository secrets:
https://github.com/hmanzur/actions-set-secret

dperaza4dustbit:
> 

Hum... not sure we want this @mmulholla that will mean we need to provide the secret for the cluster kubeadmin which as we know expires. Running a job to reset secret in between Clusters deployment I think is an overkill, the cluster deployment seems to be very manual now. What I would recommend will be to try to automated the cluster creation and then include this step to create SA and update secrets with server and token. What do you think?

mmulholla:
OK, I probably misunderstand this a bit. Lets discuss later. This LGTM for now,

mmulholla (approved):
LGTM

PR #80, opened 2021-11-30
Added sha value mismatch and invalid url test scenarios

tisutisu:
This PR will cover below stories: 
https://issues.redhat.com/browse/HELM-265
https://issues.redhat.com/browse/HELM-280

### DIFF
--- a/tests/functional/features/invalid_url_in_the_report.feature
+++ b/tests/functional/features/invalid_url_in_the_report.feature
@@ -1,0 +1,26 @@
+Feature: Report contains an invalid URL
+    Partners, redhat and community users submits only report with an invalid URL
+
+    Examples:
+      | report_path               |
+      | tests/data/report.yaml    |
+    
+    Scenario Outline: A user submits a report with an invalid url
+        Given the vendor <vendor> has a valid identity as <vendor_type>
+        And report used in <report_path>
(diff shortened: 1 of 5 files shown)


mmulholla (commented):
I would like to understand if this test could follow a similar pattern to that I used for chart-verifier test. See: https://github.com/redhat-certification/chart-verifier/pull/220

Another alternative is to add a json file for each chart which defines the message it expects. Then provide the location of the json file as a parameter. This way the code can be used for other issues in the report.

tisutisu:
> I would like to understand if this test could follow a similar pattern to that I used for chart-verifier test. See: [redhat-certification/chart-verifier#220](https://github.com/redhat-certification/chart-verifier/pull/220)
> 
> Another alternative is to add a json file for each chart which defines the message it expects. Then provide the location of the json file as a parameter. This way the code can be used for other issues in the report.

Added this change, as discussed.

mmulholla (approved):
LGTM - nice work!

zonggen:
Change LGTM, please rebase to latest main so that parallel testing has no impact.

tisutisu:
> Change LGTM, please rebase to latest main so that parallel testing has no impact.

rebase to main, now all the test are passing.

PR #81, opened 2021-12-01
Added test scenario: PR includes a non chart related file

tisutisu:
This PR will cover story: https://issues.redhat.com/browse/HELM-266


mmulholla (commented):
With every new test we are adding yet more duplicate code which we need to get away form. It is possible to reuse steps: see: https://pytest-bdd.readthedocs.io/en/5.0.0/#reusing-steps

tisutisu:
> With every new test we are adding yet more duplicate code which we need to get away from taht. It is possible to reuse steps: see: https://pytest-bdd.readthedocs.io/en/5.0.0/#reusing-steps

Addressed this comment in new PR : [#83](https://github.com/openshift-helm-charts/development/pull/83)

mmulholla (approved):
LGTM

PR #82, opened 2021-12-01
tests: check lock before modifying git config

zonggen:
https://github.com/openshift-helm-charts/development/runs/4377741765?check_suite_focus=true
https://github.com/openshift-helm-charts/development/runs/4377834190?check_suite_focus=true
https://github.com/openshift-helm-charts/development/runs/4378053172?check_suite_focus=true

For reference, here's the error message after using multi-thread testing.
`
E   OSError: Lock for file '/home/runner/work/development/development/.git/config' did already exist, delete '/home/runner/work/development/development/.git/config.lock' in case the lock is illegal
`

### DIFF
--- a/tests/functional/utils/chart_certification.py
+++ b/tests/functional/utils/chart_certification.py
@@ -51,8 +51,11 @@ vendor:
         username (str): git username to set
         email (str): git email to set
         """
-        repo.config_writer().set_value("user", "name", username).release()
-        repo.config_writer().set_value("user", "email", email).release()
+        git_path = repo.git.rev_parse("--show-toplevel")
+        # make sure only one process is modifying the git config
+        if not pathlib.Path(f'{git_path}/.git/config.lock').exists():
+            repo.config_writer().set_value("user", "name", username).release()
+            repo.config_writer().set_value("user", "email", email).release()
 
     def get_bot_name_and_token(self):
         bot_name = os.environ.get("BOT_NAME")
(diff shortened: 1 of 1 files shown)


tests/functional/utils/chart_certification.py
```diff
@@ -51,8 +51,11 @@ def set_git_username_email(self, repo, username, email):
         username (str): git username to set
         email (str): git email to set
         """
-        repo.config_writer().set_value("user", "name", username).release()
-        repo.config_writer().set_value("user", "email", email).release()
+        git_path = repo.git.rev_parse("--show-toplevel")
+        # make sure only one process is modifying the git config
+        if not pathlib.Path(f'{git_path}/.git/config.lock').exists():
```
tisutisu, line 56: I think even with this check there is a possibility that two pytest thread can check this condition at the same time and for both the thread this condition is true and try to execute the next portion of code. 
One approach I can think of is implement the similar mechanism as config_writer() using to synchronize the update of config, implement some .lock file to synchronize this portion of code.

tisutisu (changes requested):

tests/functional/utils/chart_certification.py
zonggen: The `config.lock` file was created by git functioning as a mutex lock, and I agree it might not be atomic. Don't think we can add a lock _after_ the contention point because we will be already failing at that point. Did you encounter failure case for this PR?

tisutisu: yes I encountered the same error when running with this change as well.

PR #85, opened 2021-12-08
add helm chart releaser binary

mmulholla:
Adding the 1.3.0 version helm chart verifier tgz for use in the certification process. Currently we download it during the certification process which can occasionally fail to download:

current code is:
```
        curl -L -o cr.tar.gz https://github.com/helm/chart-releaser/releases/download/v1.2.0/chart-releaser_1.2.0_linux_amd64.tar.gz
        tar zxvf cr.tar.gz
        sudo cp -f cr /usr/local/bin/cr
```

this will be updated in a separate PR (to prevent adding a bad tar as part of a PR and running it in the same PR):
```
          tar zxvf ./scripts/dependencies/helm-chart-releaser/chart-releaser_1.3.0_linux_amd64.tar.gz
          sudo cp -f cr /usr/local/bin/cr
```

By adding the binary under the script directory so the binary is considered CI changes so only folks in our OWNERS file can update the binary.

PR #86, opened 2021-12-08
add use of downloaded helm chart releaser in the workflow

mmulholla:
Update workflow to use the downloaded chart releaser in https://github.com/openshift-helm-charts/development/pull/85


mmulholla:
This test PR shows the code working: https://github.com/openshift-helm-charts/sandbox/pull/6013

mmulholla:
Tested on my fork and found both chart releaser 1.2.0 and 1.3.0 created a release correctly, so I think this change is good. However, we should wait for a clean test run before merging once the tests have been fixed.

zonggen (approved):
LGTM!

PR #92, opened 2021-12-28
Added test scenario: report in json format

tisutisu:
This PR will resolve story: https://issues.redhat.com/browse/HELM-287


tests/functional/utils/chart_certification.py
```diff
@@ -466,7 +466,7 @@ def process_report(self, update_chart_sha=False, update_url=False, url=None):
             values = {'repository': self.secrets.test_repo,
                     'branch': self.secrets.base_branch}
             content = Template(tmpl).substitute(values)
-            report_path = f'{self.chart_directory}/{self.secrets.chart_version}/report.yaml'
+            report_path = f'{self.chart_directory}/{self.secrets.chart_version}/' + self.secrets.test_report.split('/')[-1]
```
zonggen, line 469: can the last opearand `self.secrets.test_report.split('/')[-1]` part of the format string?

tisutisu: No, it is separate. In the last operand we are extracting the filename i.e. report.yaml or report.json(applicable for this json formatted report scenario) from the self.secrets.test_report variable. We cannot add string operations inside a format string.

zonggen (approved):
LGTM

PR #94, opened 2022-01-12
added certification flow smoke suite, nightly test run workflow, release workflow to run full set of tests and readme

tisutisu:
This PR will cover story: https://issues.redhat.com/browse/HELM-291

1. created smoke test suite (a subset of the all tests) which will during PR submission with workflow changes.
2. Added github workflow to schedule overnight runs for complete set of tests.
3. README.md file added, which describes how to determine if future tests are included.
4. changes added in the release workflow to run all the test


.github/workflows/release.yml
```diff
@@ -116,6 +116,17 @@ jobs:
           cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
           cd scripts && ../ve1/bin/python3 setup.py install && cd ..
 
+      - name: Run All Tests
```
mmulholla, line 119: we don't want this here,  It needs to be in test.yml. Currently in test.yaml we decide not to run any tests for a release and then approve and merge the release PR. That needs to change so that it now does run tests for a release and does not merge and release if tests fail. The test then need to be the full bucket. This flow is about creating PR's to copy charts from the chart repo to dev repo and copy flows to charts repo.

.github/workflows/test.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_BODY: "Test triggered by ${{ github.event.pull_request.html_url }}."
         run: |
-          ve1/bin/pytest tests/ --log-cli-level=WARNING --ignore=tests/functional/step_defs/test_submitted_charts.py --tb=short
+          ve1/bin/pytest tests/functional/step_defs/test_smoke_scenarios.py --log-cli-level=WARNING --tb=short
```
mmulholla, line 85: It is here we need to run full test for the release process.

Also, a new requirement but we should only do smoke test to if PR is in development repo. If we are updating CI workflow in charts repo we should also run the full bucket.

mmulholla (commented):
Excellent work! Just need to move where we do the full test.

.github/workflows/release.yml
mmulholla: My bad, we do want it here but would good to have a few steps earlier to avoid the unnecessary checkout and setting up PR scripts again.

mmulholla (commented):
Also need to run full test in test.yaml for the development to charts PR so if ```steps.check_if_release_pr.outputs.charts_release_branch == 'true'```

To do this you will need to reorder some steps.

.github/workflows/release.yml
tisutisu: Updated workflow to addressed this comment.

.github/workflows/test.yml
tisutisu: Addressed this new requirement as well.

mmulholla (commented):
I ran a dummy release to check with release workflow and got some ugly failures running the tests. It may be because the release does not not include any CI changes. We may need to include release/release_info.json and a CI change - not sure.

.github/workflows/test.yml
```diff
@@ -76,13 +76,23 @@ jobs:
...
         env:
+          GITHUB_REPO: ${{ github.repository }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           BOT_NAME: ${{ secrets.BOT_NAME }}
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_BODY: "Test triggered by ${{ github.event.pull_request.html_url }}."
         run: |
-          ve1/bin/pytest tests/ --log-cli-level=WARNING --ignore=tests/functional/step_defs/test_submitted_charts.py --tb=short
+          if [ ${GITHUB_REPO} == 'openshift-helm-charts/charts' ]; then
```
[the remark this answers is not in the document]
tisutisu: changed the approach now. using below check to differentiate between CI change only and dev to charts release PR:
```
steps.check_if_release_pr.outputs.charts_release_branch  == 'true'
```

.github/workflows/test.yml
```diff
@@ -76,13 +76,23 @@ jobs:
...
+            echo "Charts repo is targetted, so running full tests"
+            ve1/bin/pytest tests/ --log-cli-level=WARNING --ignore=tests/functional/step_defs/test_smoke_scenarios.py --ignore=tests/functional/step_defs/test_submitted_charts.py --tb=short
+          elif [ ${GITHUB_REPO} == 'openshift-helm-charts/development' ]; then
```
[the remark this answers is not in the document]
tisutisu: See my comment above. Will be addressed by the same.

.github/workflows/test.yml
```diff
@@ -76,13 +76,23 @@ jobs:
...
+            echo "Developement repo is targetted, so running only smoke tests"
+            ve1/bin/pytest tests/functional/step_defs/test_smoke_scenarios.py --log-cli-level=WARNING --tb=short
+          else
+            echo "Unexpected condition: exiting"
```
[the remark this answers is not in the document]
tisutisu: see my above comment. will not require this now.

tisutisu:
Addressed the review comments. Also tested the change in my own fork. See below:

CI Change test: https://github.com/tisutisu/development/pull/11

[Release Test]
RELEASE_PR: https://github.com/tisutisu/development/pull/9
CHARTS_TO_DEV_PR: https://github.com/tisutisu/development/pull/10
DEV_TO_CHARTS: It was not created due to no changes required

mmulholla (approved):
Great work! I approve.

.github/workflows/nightly_test.yml
```diff
@@ -0,0 +1,47 @@
+name: Run Nightly Test
```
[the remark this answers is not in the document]
tisutisu: Updated this PR with the same. Added this file to ignore list in release/release_info.json

PR #95, opened 2022-01-21
chart-verifier: use GitHub Action to download cv binary

zonggen:
Uses https://github.com/redhat-actions/openshift-tools-installer to
download and chart-verifier binary. Also removes dependencies to docker
binaries.

Closes: https://issues.redhat.com/browse/HELM-273

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ jobs:
   chart-certification:
     name: Chart Certification
     runs-on: ubuntu-20.04
-    env:
-      VERIFIER_IMAGE: quay.io/redhat-certification/chart-verifier:main
     if: |
       github.event.pull_request.draft == false &&
       (github.event.action != 'labeled' || github.event.label.name == 'force-publish')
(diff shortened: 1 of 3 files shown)


zonggen:
Ready for reivews, some tests are failing because sandbox is running the `build.yaml` in sandbox's main branch.

mmulholla (commented):
I am having second thoughts on this:
- I expected the github action to generate the report but is is not. 
- We can't use latest in the development branch, we need main to be able to test with new chart verifier updates, and you can't use the github action with main.
- We should not be treating the certification flow as a test.

If we really want to test the github action we should have a specific test flow in the chart-verifier, not our certification flow.

Also be aware the chart verifier builds and tests the binary on every PR.

.github/workflows/build.yml
```diff
@@ -162,6 +160,18 @@ jobs:
...
+        id: install-chart-verifier
+        if: ${{ steps.check_build_required.outputs.run-build == 'true' }}
+        uses: redhat-actions/openshift-tools-installer@v1
+        with:
+          # Search through these projects' GitHub releases, instead of the OpenShift Mirror.
+          # https://docs.github.com/en/actions/reference/authentication-in-a-workflow
+          source: "github"
+          # Using GitHub token from the github context
+          github_pat: ${{ secrets.GITHUB_TOKEN }}
+          chart-verifier: "latest"
```
mmulholla, line 173: For development we need to use "main" which unfortunately does not have a binary. Don't think we can use it in this form. A test of the github action is better in the chart verifier builds as a separate workflow. 
Also we do not want the chart-certifcation flow as a testing ground for the action.

mmulholla (commented):
I don't think this should be merged.

PR #96, opened 2022-01-31
Update for chart-verifier report to stdout

mmulholla:
Modify flows to expect chart verifier report to be output to stdout.


dperaza4dustbit (approved):
lgtm

PR #110, opened 2022-02-17
add provider delivery control and tests

mmulholla:
This PR includes:

- Renamed ```sanitycheck``` to ```checkpr```. Use of ```sanity``` disrespects folks with mental health issues.
      - also uses label ```content-ok``` to replace ```sanity-ok``` so those labels need to be defined in the repositories. 
      - see: <https://gist.github.com/seanmhanson/fe370c2d8bd2b3228680e38899baf5cc#:~:text=The%20phrase%20sanity%20check%20is,to%20discriminate%20against%20these%20people.>
- Added code to write index file content to ```gh-pages``` branch and file  ```unpublished-certified-charts.yaml``` if:
     - the report and the OWNERS file both  indicate provider delivery control
     - the PR is report only
     - all required tests in the report pass
     - error situations:
          - OWNERS file and report are inconsistent in the provider delivery control setting
          - OWNERS file and report indicate provider delivery control but the PR is not report only.  
- Added pyTests have to check the working case and each error case
    -  the working case test has also been added to smoke tests - it is report only so relatively quick.
-  Added logic to run a full test when a PR includes changes or additions to the full test.

Note: have not limited the provider controlled delivery to partners, seem equally valid for Red Hat and community

Note: This PR will just run the smoke tests, rest assured I have run twice with full test suite and everything passed.


scripts/src/checkprcontent/checkpr.py
```diff
@@ -10,10 +10,73 @@
...
+            msg = f"[ERROR] OWNERS file and/or report indicate provider controlled delivery but pull request is not report only."
+            print(msg)
+            print(f"::set-output name=pr-content-error-message::{msg}")
+            sys.exit(1)
+    elif report_in_pr:
+        if report_provider_delivery and owner_provider_delivery:
+            print(f"[INFO] providerDelivery is a go")
+            print(f"::set-output name=providerDelivery::True")
+        elif report_provider_delivery:
+            msg = f"[ERROR] Report indicates provider controlled delivery but OWNERS file does not."
```
dperaza4dustbit, line 64: I would only use the f-strings when I'm parametrizing the string.

dperaza4dustbit (approved):
One comment to improve

scripts/src/checkprcontent/checkpr.py
mmulholla: will update later.

PR #112, opened 2022-03-01
fix bug in ProviderDelivery

mmulholla:
Two issues.
1. When deciding which index file to use the test checked for ```os.environ["PROVIDER_DELIVERY"]``` as a binary, but it is a string. As a result everything went to Index.yaml.
2. The test case missed out the call to check the content of the new index file.


dperaza4dustbit (approved):
lgtm

PR #113, opened 2022-03-01
print errors from verifier and ensure delete namespace is called

mmulholla:
chartprreview.py:
- Now that the verifier outputs errors to stderr - I added code to check for stderr and output the content.
- Removed one print out of the report as it will be printed later.

build.yml:
Delete Namespace was not being called as it was checking for a non existent output. Fixed it.

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -340,9 +340,13 @@ def generate_verify_report(directory, category, organization, chart, version):
 
     stdout = out.stdout.decode("utf-8")
     report_path = "report.yaml"
-    print("[INFO] report:\n", stdout)
     with open(report_path, "w") as fd:
         fd.write(stdout)
+
+    if len(out.stderr) >0 :
+        print(f'[WARNING] Errors creating report:\n{out.stderr.decode("utf-8")}')
+
+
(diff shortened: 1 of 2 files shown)


scripts/src/chartprreview/chartprreview.py
```diff
@@ -340,10 +340,14 @@ def generate_verify_report(directory, category, organization, chart, version):
 
     stdout = out.stdout.decode("utf-8")
     report_path = "report.yaml"
-    print("[INFO] report:\n", stdout)
     with open(report_path, "w") as fd:
         fd.write(stdout)
 
+    if len(out.stderr) >0 :
+        print(f'[WARNING] Errors creating report:\n{out.stderr.decode("utf-8")}')
+
```
tisutisu, line 348: In case we got stderr when running the docker cmd, do still want to continue? I mean in case of failure report.yaml will contain empty file. Don't we have an error here instead of a warning?

mmulholla: No, the errors typically come out if chart testing fails which is a lot, but you still get a report. If we got errors and not report that would likely be a bug in the script.

dperaza4dustbit (approved):
lgtm

PR #115, opened 2022-03-07
initialise python in metrics flow

mmulholla:
Update flow to fully initialize python as per our other flow.

Also added workflow dis[patch so we can run (and test) on demand.


dperaza4dustbit (approved):
lgtm

PR #116, opened 2022-03-08
Add slack notification if tests are run

mmulholla:
Add slack notification if a test is run either due to a OCP version update or chart verifier version update.


dperaza4dustbit (approved):
lgtm

PR #121, opened 2022-03-10
Use python library for analytics

mmulholla:
Now using python analytics library i stead using HTTP requests - make the code a lot simpler and fixes problem where HTTP code was not uploading.

metrics scripts also includes code ready for pull request metrics.

also added code so that data uploaded now looks like this: 

analytics.track('HashiCorp', 'vault downloads', {
    'hashicorp-vault-0.14.0.tgz': 78
})

this makes the data more granular

This adds a new file index.py most of which was taken from @dperaza4dustbit python script which reads the index file.

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Release Metrics
         id: release_metrics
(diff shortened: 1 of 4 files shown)


dperaza4dustbit (approved):
lgtm

PR #125, opened 2022-03-23
add PR metrics and update release metrics

mmulholla:
- Added metrics each time a chart PR runs. Added at then end of the workflow so does not affect merge if it fails for any reason.
- Modified release metrics to log a single metric for all versions of the same chart. This should prevent woopra showing lots of zeros.
- Added the concept of a test segment write key and  production write key read from secrets
    - SEGMENT_WRITE_KEY - production key
    - SEGMENT_TEST_WRITE_KEY - test key
- Added a workflow-dispatch for the metrics workflow, takes a single boolean input to se which segment write key to use.
     - dispatch workflow can be run form forks.

This PR was created for: https://issues.redhat.com/browse/HELM-302


dperaza4dustbit (approved):
lgtm

PR #127, opened 2022-04-12
Use chart verifier and oc actions

mmulholla:
Have updated the flow to use the chart verifier action to create the report and the report summary and as a result the code required to generate those is no longer needed.

Also updated the flow to use existing actions to perform oc install and oc login helping to simplify the workflow.

There were knock on affects in may parts of the scripts which support the flow.


dperaza4dustbit (approved):
lgtm

PR #132, opened 2022-05-31
require a package digest for PDC

mmulholla:
Changes made:

if submission is for provider deliver control and package digest is not set , fail the PR
if provider delivery control is set do not check the chart uri in the report.
include a check of the digest as part of checking the chart uri so if there is an issue fail before merge.

added a test for provider deliver control when the report  does not include a package digest.
added package digest to reports used in tests.

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -156,6 +156,8 @@ def check_url(directory, report_path):
         msgs.append(str(err))
         write_error_log(directory, *msgs)
 
+    verify_package_digest(chart_url,report_path)
+
 def match_name_and_version(directory, category, organization, chart, version,generated_report_path):
     print("[INFO] Check chart has same name and version as directory structure. %s, %s, %s" % (organization, chart, version))
     submitted_report_path = os.path.join("charts", category, organization, chart, version, "report.yaml")
@@ -300,7 +302,23 @@ def check_report_success(directory, api_url, report_path, report_info_path, vers
             write_error_log(directory, msg)
             sys.exit(1)
 
+def verify_package_digest(url,report):
+    print("[INFO] check package digest.")
 
+    response = requests.get(url, allow_redirects=True)
+    target_digest = hashlib.sha256(response.content).hexdigest()
+
+    found,report_data = verifier_report.get_report_data(report)
+    if found:
+        pkg_digest = verifier_report.get_package_digest(report_data)
+
+    if target_digest:
+        if pkg_digest and pkg_digest != target_digest:
+            # Digest was passed and computed but differ
+            raise Exception("Found an integrity issue. SHA256 digest passed does not match SHA256 digest computed.")
+    elif not pkg_digest:
+        # Digest was not passed and could not be computed
+        raise Exception("Was unable to compute SHA256 digest, please ensure chart url points to a chart package.")
 
 
 def main():
@@ -331,6 +349,7 @@ def main():
     report_generated = os.environ.get("REPORT_GENERATED")
     generated_report_path = os.environ.get("GENERATED_REPORT_PATH")
     generated_report_info_path =  os.environ.get("REPORT_SUMMARY_PATH")
+    provider_delivery = os.environ.get("PROVIDER_DELIVERY")
 
     if os.path.exists(submitted_report_path):
         print("[INFO] Report exists: ", submitted_report_path)
@@ -339,7 +358,7 @@ def main():
         report_info_path = ""
         if report_generated and report_generated == "True":
             match_checksum(args.directory,generated_report_info_path, category, organization, chart, version)
-        else:
+        elif not provider_delivery or provider_delivery == "False":
             check_url(args.directory, report_path)
     else:
         print("[INFO] Report does not exist: ", submitted_report_path)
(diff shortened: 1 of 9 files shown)


scripts/src/chartprreview/chartprreview.py
```diff
@@ -300,7 +302,23 @@ def check_report_success(directory, api_url, report_path, report_info_path, vers
...
+    if found:
+        pkg_digest = verifier_report.get_package_digest(report_data)
+
+    if target_digest:
+        if pkg_digest and pkg_digest != target_digest:
+            # Digest was passed and computed but differ
+            raise Exception("Found an integrity issue. SHA256 digest passed does not match SHA256 digest computed.")
+    elif not pkg_digest:
+        # Digest was not passed and could not be computed
+        raise Exception("Was unable to compute SHA256 digest, please ensure chart url points to a chart package.")
```
isutton, line 322: I'd simplify the flow a little bit, including an additional exception when report data can't be found in the report.

```python
def verify_package_digest(url, report):
    print("[INFO] check package digest.")

    response = requests.get(url, allow_redirects=True)
    target_digest = hashlib.sha256(response.content).hexdigest()

    if not target_digest:
        raise Exception("Was unable to compute SHA256 digest, please ensure chart url points to a chart package.")

    found, report_data = verifier_report.get_report_data(report)
    if not found:
        raise Exception("Was unable to collect report data, can't compute SHA256 digest without it.")

    pkg_digest = verifier_report.get_package_digest(report_data)

    if pkg_digest != target_digest:
        raise Exception("Found an integrity issue. SHA256 digest passed does not match SHA256 digest computed.")
```

isutton (changes requested):

scripts/src/chartprreview/chartprreview.py
```diff
@@ -339,7 +358,7 @@ def main():
         report_info_path = ""
         if report_generated and report_generated == "True":
             match_checksum(args.directory,generated_report_info_path, category, organization, chart, version)
-        else:
+        elif not provider_delivery or provider_delivery == "False":
```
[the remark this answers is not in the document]
mmulholla: Yes and yes, this logic is running when a report was submitted in the PR. In this case we may have also generated a report to match checksum, but in provider delivery this would not be the case. Before at this stage we would check for a valid chart url in the report, but for provider delivery control a the chart url will not be set.

scripts/src/chartprreview/chartprreview.py
isutton: I've added the changes you've pointed out and an explicit check for when the url can't be retrieved, together with documentation regarding the conditions you've mentioned:

```python
def verify_package_digest(url, report):
    """
    Verifies whether the artifact found at given `url` and the digest contained 
    in the given `report`. It raises an exception when:

    - The given report doesn't contain valid report data.
    - The report data contained in the given report doesn't contain a valid package digest
    - The given url's digest is different than the digest available in report data.
    """
    print("[INFO] check package digest.")

    found, report_data = verifier_report.get_report_data(report)
    
    if not found:
        raise Exception("Was unable to collect report data, can't compute SHA256 digest without it.")

    # pkg_digest can be None
    pkg_digest = verifier_report.get_package_digest(report_data)

    # raises exception if pkg_digest is None or ""
    if not pkg_digest:
        raise Exception("Was unable to collect package digest, can't compute SHA256 digest without it.)

    # download the chart
    response = requests.get(url, allow_redirects=True)

    if response.status_code == 200:
        # calculate the chart's digest from the response contents if the given url is reachable
        target_digest = hashlib.sha256(response.content).hexdigest()

        # when both pkg_digest and target_digest are set, and they're not equal raise an exception.
        if target_digest and pkg_digest != target_digest:
            raise Exception("Found an integrity issue. SHA256 digest passed does not match SHA256 digest computed.")
    else:
        # no-op when the url's SHA256 can't be computed but the report contains a digest.
        pass
```

scripts/src/chartprreview/chartprreview.py
```diff
@@ -331,6 +349,7 @@ def main():
     report_generated = os.environ.get("REPORT_GENERATED")
     generated_report_path = os.environ.get("GENERATED_REPORT_PATH")
     generated_report_info_path =  os.environ.get("REPORT_SUMMARY_PATH")
+    provider_delivery = os.environ.get("PROVIDER_DELIVERY")
```
isutton, line 352: I see a lot of `if not ... or ... == "False":` that could be served better by the `environs` Python package, which gives a construct that turns the string value into a boolean:

```python
import environs
from environs import Env

env = Env()
provider_delivery = env.bool("PROVIDER_DELIVERY", False)
```

scripts/src/chartprreview/chartprreview.py
mmulholla: Not trying to introduce any new errors here compared to previous code later in the process which is from where I copied the logic: https://github.com/openshift-helm-charts/development/blob/a8c241f353eb332a02cfbb66842fde96766cf815/scripts/src/chartrepomanager/chartrepomanager.py#L170

Not getting a pkg_digest is only an error if we also cannot get a package digest. With this in mind I think the original logic is fine except for not checking for a 200 response which I now added.

isutton (approved):
lgtm

PR #134, opened 2022-06-09
version_check changes needed for release

mmulholla:
The release build failed because the script was expecting to update file schedule.yml which is now renamed to version_check.yml. Further even with the correct name it would have failed because the cron job invocation was incorrectly in the development version of version_check.yml so it is now removed. '

Also added a quick fix for the smoke tests to fix the provider delivery control failure (test problem)


dperaza4dustbit (approved):
lgtm

PR #137, opened 2022-06-15
Exit early if more that one chart in a PR

mmulholla:
Improved error message and added an exit after detecting multiple charts in a PR.

Test framework does not support adding a multi-chart PR see: https://issues.redhat.com/browse/HELM-368
Manual test was performed see: https://github.com/mmulholla/development/pull/286

This PR addresses bug: https://issues.redhat.com/browse/HELM-365


dperaza4dustbit (approved):
lgtm

PR #140, opened 2022-06-23
Ensure label is added and add test

mmulholla:
The problem happens when the chart version includes a "-" character, and is nothing to do with provider delivery. This happened because checkpr.py uses a pattern match which includes a "-" in the version when checking pr content is valid. However when creating data for the label step prartifact.py used a patten which did not include a "-" in the version and as a result did not find any files in the PR.

Included in the fix:

- Added a method to checkpr.py to return the pattern match strings which are now used in checkpr.py and prartifcat.py. As a result we have a single source for the pattern match strings which allow for "-" in the version.
- Added code to the test framework to check for the "content-ok" and "authorized-request" labels for all tests where the PR is merged.
- Added a test where the chart version include a "-" character and is expected to merge.
   - Created report files for each test.
- Fixed an issues in the test framework created by the new test where the chart directory used by the pull request was set too soon and relied on all tests using the same report.


scripts/src/checkprcontent/checkpr.py
```diff
@@ -84,6 +84,11 @@ def check_provider_delivery(report_in_pr,num_files_in_pr,report_file_match):
         print(f"::set-output name=providerDelivery::False")
         print(f"[INFO] providerDelivery is a no-go")
 
+def get_file_match_compiled_patterns():
```
isutton, line 87: A docstring explaining what sort of inputs those patterns ingest might clarify the regular expressions below.

scripts/src/pullrequest/prartifact.py
```diff
@@ -25,9 +27,11 @@ def get_modified_charts(api_url):
 
 def save_metadata(directory, vendor_label, chart, number):
     with open(os.path.join(directory, "vendor"), "w") as fd:
+        print(f"add {directory}/vendor as {vendor_label}")
         fd.write(vendor_label)
 
     with open(os.path.join(directory, "chart"), "w") as fd:
+        print(f"add {directory}/chart as {chart}")
         fd.write(chart)
```
isutton, line 35: Are those `print` statements left overs?

mmulholla: Yes, but I left them in because they only come out when there is a failure and can be useful. We have way too few of these in the flows.

scripts/src/checkprcontent/checkpr.py
isutton: Yes, and also some description of the captures it might provide, for example:

```python
"""Return a tuple of patterns, where  the first can be used to match a chart PR using any URL, and the second a pattern to match a valid report file within a PR URL.

Both patterns capture ?, ? and ? from the URL.

Examples of valid URLs are:

charts/.../
charts/..../report.yaml
```

mmulholla: @isutton I added a docstring based on your input. Hope you like it!

isutton (approved):
LGTM.

PR #144, opened 2022-07-21
Fix bash error and update workflow to enable better testing

mmulholla:
https://issues.redhat.com/browse/HELM-383

Fixed formatting error in the bash script part of a step. 

The problem happened because some of the workflow only ran in the schedule job. As a result I also updated the flow to:
1. enable the entire workflow to be tested by adding a new update-version file attribute to the  manual run.
2. Enable the flow to modify a local copy of the software-version file for a manual workflow run not in the charts repo.  

Finally I update the two jobs to run synchronously to prevent a conflict when updating the software version file.

As a result the workflow has been fully tested.

### DIFF
--- a/.github/workflows/version_check.yml
+++ b/.github/workflows/version_check.yml
@@ -7,9 +7,13 @@ on:
       # 2. Runs this workflow whether or not a version change has occurred
       # 3. By default dperaza and mmulholla are tagged in any issues raised.
       dry-run:
-        description: "Dry Run? (Run tests but create issues in sandbox) {true,false}"
+        description: "Dry Run? (Unconditionally run tests and create issues in sandbox) {true,false}"
         required: true
         default: "true"
+      update-version:
+        description: "Dry run also checks and updates software-version file if not charts repository"
+        required: true
+        default: "false"
       vendor-type:
         description: "Vendor type {all,partner,redhat,community}"
         required: true
@@ -29,13 +33,27 @@ jobs:
         run: |
           echo "GITHUB_EVENT_NAME : $GITHUB_EVENT_NAME"
           echo "GITHUB_REPOSITORY : $GITHUB_REPOSITORY"
+          echo "dry-run : ${{ github.event.inputs.dry-run }}"
+          echo "update-version : ${{ github.event.inputs.update-version }}"
           if [ $GITHUB_EVENT_NAME == 'workflow_dispatch' ]; then
             echo '::set-output name=run-job::true'
+            if [ "${{ github.event.inputs.dry-run }}" == "true" ]; then
+                if [[ "${{ github.event.inputs.update-version }}" == "true"  && $GITHUB_REPOSITORY != "openshift-helm-charts/charts" ]]; then
+                  echo '::set-output name=check-version::true'
+                else
+                  echo '::set-output name=check-version::false'
+                fi
+            else
+                echo '::set-output name=check-version::true'
+            fi
           elif [ $GITHUB_REPOSITORY == "openshift-helm-charts/charts" ]; then
             echo '::set-output name=run-job::true'
+            echo '::set-output name=check-version::true'
           else
             echo '::set-output name=run-job::false'
+            echo '::set-output name=check-version::false'
           fi
+
 
       - name: Install oc
         if: steps.check_repo.outputs.run-job == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/version_check.yml
```diff
@@ -55,56 +73,67 @@ jobs:
         id: get_curr_ocp_version
         run: |
           OCP_VERSION=$(./oc version -o json | jq '.openshiftVersion')
+          OCP_VERSION=$(sed -e 's/^"//' -e 's/"$//' <<< $OCP_VERSION)
```
tisutisu, line 76: I think for better readability of the code, we can combine in a single statement 
` OCP_VERSION=$(./oc version -o json | jq '.openshiftVersion' | sed -e 's/^"//' -e 's/"$//' ) `

.github/workflows/version_check.yml
```diff
@@ -355,13 +407,14 @@ jobs:
...
-        if: always() && github.event_name == 'schedule' && steps.compare_cv_versions.outputs.run_tests == 'true'
+        if: always() && github.event_name == 'schedule' && steps.check_test.outputs.run_tests == 'true'
         uses: archive/github-actions-slack@v2.0.0
         with:
           slack-bot-user-oauth-access-token: ${{ secrets.SLACK_BOT_USER_OAUTH_ACCESS_TOKEN }}
           slack-channel: C02979BDUPL
           slack-text: ${{ steps.run-schedule-tests.conclusion }}! Nightly run after chartverifier version update detected. See '${{github.server_url}}/${{github.repository}}/actions/runs/${{github.run_id}}'
 
       - name: Result from "Send Message to slack channel"
+        if: always() && github.event_name == 'schedule' && steps.check_tes.outputs.run_tests == 'true'
```
tisutisu, line 418: there is an typo here for `check_test`

.github/workflows/version_check.yml
```diff
@@ -55,56 +73,67 @@ jobs:
...
         shell: bash
 
       - name: Checkout software-version branch
-        if: steps.check_repo.outputs.run-job == 'true'
+        if: steps.check_repo.outputs.check-version == 'true'
         uses: actions/checkout@v2
         with:
           ref: "software-version"
-          repository: "openshift-helm-charts/charts"
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
tisutisu, line 86: Will it be running with event pull_request also?

.github/workflows/version_check.yml
mmulholla: NIce spot. thx

.github/workflows/version_check.yml
mmulholla: Good sport. update to use ${{ github.repository }}. Interestingly the above did not cause failures because it evaluates to null and the default value works.

.github/workflows/version_check.yml
```diff
@@ -355,13 +407,14 @@ jobs:
 
       - name: Send message to slack channel
         id: notify
-        if: always() && github.event_name == 'schedule' && steps.compare_cv_versions.outputs.run_tests == 'true'
+        if: always() && github.event_name == 'schedule' && steps.check_test.outputs.run_tests == 'true'
         uses: archive/github-actions-slack@v2.0.0
         with:
           slack-bot-user-oauth-access-token: ${{ secrets.SLACK_BOT_USER_OAUTH_ACCESS_TOKEN }}
           slack-channel: C02979BDUPL
           slack-text: ${{ steps.run-schedule-tests.conclusion }}! Nightly run after chartverifier version update detected. See '${{github.server_url}}/${{github.repository}}/actions/runs/${{github.run_id}}'
```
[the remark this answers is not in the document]
mmulholla: I improved the messages to include the new version.

isutton (commented):
LGTM.

tisutisu (approved):
LGTM. Only concern I have is test failure on this PR. Not sure if it is due to this PR is not rebased with the behave migration changes which we merged recently. Anyway we can debug that in other PR also.

.github/workflows/version_check.yml
mmulholla: Prefer to leave as-is. Do not like single line multi-commands because it one fails it is more difficult to find the culprit.

PR #147, opened 2022-08-02
check OCP version support in index.yaml

mmulholla:
https://issues.redhat.com/browse/HELM-358

Updated version check workflow to check that the OCP Versions or kubeVersion of all charts  (latest version only) in index.yaml include the OCP version which caused the workflow to run.  

For this needed a mew message when opening a PR, for example:
```
FYI @mmulholla, we checked the OCP versions supported by ibm ibm-oms-ent-prod 6.0.2 because the workflow now supports OpenShift version 4.10.25. We have found that chart ibm-oms-ent-prod 6.0.2 kubeVersion 1.16.0-0 (OCP: 4.3) does not include latest OCP version 4.10.25. Chart owners are: ['ntinvo']
``` 

also improved other messages. 

Also a fix to read the OCP version to kubeVersion map from a chart verifier file so we have a single source.

### DIFF
--- a/scripts/src/indexfile/index.py
+++ b/scripts/src/indexfile/index.py
@@ -8,14 +8,10 @@ import sys
 sys.path.append('../')
 from chartrepomanager import indexannotations
 
-def _make_http_request(method, url, body=None, params={}, headers={}, verbose=False):
-    method_map = {"get": requests.get,
-                  "post": requests.post,
-                  "put": requests.put,
-                  "delete": requests.delete,
-                  "patch": requests.patch}
-    request_method = method_map[method]
-    response = request_method(url, params=params, headers=headers, json=body)
+INDEX_FILE = "https://charts.openshift.io/index.yaml"
+
+def _make_http_request(url, body=None, params={}, headers={}, verbose=False):
+    response = requests.get(url, params=params, headers=headers, json=body)
     if verbose:
         print(json.dumps(headers, indent=4, sort_keys=True))
         print(json.dumps(body, indent=4, sort_keys=True))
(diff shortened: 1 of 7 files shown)


scripts/src/indexfile/index.py
```diff
@@ -38,5 +43,97 @@ def get_chart_info(tar_name):
     print(f"[INFO] match not found: {tar_name}")
     return "","","",""
 
+def get_charts_info():
+    chart_info_list = []
+
+    index_dct = _load_index_yaml("https://charts.openshift.io/index.yaml")
```
tisutisu, line 49: We are using the URL in two functions... It be good to move the URL to a global variable

scripts/src/indexfile/index.py
```diff
@@ -38,5 +43,97 @@ def get_chart_info(tar_name):
...
+            #print(f'   [INFO] last chart completed : {chart_in_process["name"]} {chart_in_process["version"]}')
+            latest_charts.append(chart_in_process)
+
+    return latest_charts
+
+
 if __name__ == "__main__":
-    get_chart_info("redhat-dotnet-0.0.1")
\ No newline at end of file
+    get_chart_info("redhat-dotnet-0.0.1")
```
tisutisu, line 111: It would have been good to see some unit tests for the new functions added to the modules. I see it missing for other modules as well. Anyway we can scope that as part of other story.

scripts/src/indexfile/index.py
```diff
@@ -38,5 +43,97 @@ def get_chart_info(tar_name):
     print(f"[INFO] match not found: {tar_name}")
     return "","","",""
 
+def get_charts_info():
+    chart_info_list = []
+
+    index_dct = _load_index_yaml("https://charts.openshift.io/index.yaml")
+    for entry, charts in index_dct["entries"].items():
+        for chart in charts:
+            chart_info = {}
```
tisutisu, line 52: It is functionally working fine..  but I think it would have been good to implement .. chart_info as an object of type Chart .. and defining a class `Chart` ... it makes the code more structured

tisutisu (commented):
LGTM .. suggested some good to have changes.. please check more on the comments

scripts/src/indexfile/index.py
mmulholla: Yes, lets scope it in another story. Don't have enough time to rewrite and do all the extra testing at the moment.

scripts/src/indexfile/index.py
mmulholla: For now I like to stick with functionally working fine. Don't have time to rework the code and complete all of the testing especially never having done python object before. Can plan to update in the next story.

tisutisu (approved):
LGTM

PR #159, opened 2022-08-31
Check chart version is a valid semantic version

mmulholla:
Added an early check for the version based on the directory structure to be valid semantic version and error if it is not. Current test suite does not support creating a chart in a directory with an invalid semantic version so have opened a jira story to add this: https://issues.redhat.com/browse/HELM-409
A manual test can be seen here: https://github.com/mmulholla/development/pull/309

Also added code to handle the chart verifier step ending in an error and not producing a report. This could happen, for example, if the directory structure uses a valid chart version but the chart itself includes a version which is not a valid semantic version. Added a  test for this as it was a simple addition. Note: without this fix, certification would fail but the message added to the PR said the chart had been certified.

### DIFF
--- a/scripts/src/checkprcontent/checkpr.py
+++ b/scripts/src/checkprcontent/checkpr.py
@@ -144,7 +144,6 @@ def ensure_only_chart_is_modified(api_url, repository, branch):
                     print(f"[INFO] Report found: {file_path}")
                     print("::set-output name=report-exists::true")
                     report_found = True
-                    report_path = file_path
                 if matches_found == 1:
                     pattern_match = match
                 elif pattern_match.groups() != match.groups():
--- a/scripts/src/checkprcontent/checkpr.py
+++ b/scripts/src/checkprcontent/checkpr.py
@@ -4,6 +4,7 @@ import sys
 import argparse
 
 import requests
+import semver
 import yaml
 try:
     from yaml import CLoader as Loader, CDumper as Dumper
@@ -140,9 +141,10 @@ def ensure_only_chart_is_modified(api_url, repository, branch):
             else:
                 matches_found += 1
                 if reportpattern.match(file_path):
-                    print("[INFO] Report found")
+                    print(f"[INFO] Report found: {file_path}")
                     print("::set-output name=report-exists::true")
                     report_found = True
+                    report_path = file_path
                 if matches_found == 1:
                     pattern_match = match
                 elif pattern_match.groups() != match.groups():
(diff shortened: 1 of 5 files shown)


scripts/src/checkprcontent/checkpr.py
```diff
@@ -140,9 +141,10 @@ def ensure_only_chart_is_modified(api_url, repository, branch):
             else:
                 matches_found += 1
                 if reportpattern.match(file_path):
-                    print("[INFO] Report found")
+                    print(f"[INFO] Report found: {file_path}")
                     print("::set-output name=report-exists::true")
                     report_found = True
+                    report_path = file_path
```
tisutisu, line 147: I think `report_path` is not needed here. We are not using it anywhere after defining.

tests/functional/behave_features/HC-14_user_submits_chart_with_errors.feature
```diff
@@ -37,7 +37,8 @@ Feature: Chart submission with errors
     Examples:
       | vendor_type  | vendor    | chart_path                     | message                               | bad_version | 
       | partners     | hashicorp | tests/data/vault-0.17.0.tgz    | doesn't match the directory structure | 9.9.9       |
-    
+      | partners     | hashicorp | tests/data/vault-0.17.0.tgz    | The chart verifier returned an error  | abc-9.9.9       |
+
```
tisutisu, line 41: Is this test resolving https://issues.redhat.com/browse/HELM-409 or do we need to any other test

scripts/src/report/verifier_report.py
```diff
@@ -84,6 +84,15 @@ def get_provider_delivery(report_data):
         pass
     return provider_delivery
 
+def get_chart_version(report_data):
```
tisutisu, line 87: Are we using this function anywhere in the workflow?

tisutisu (commented):
Overall LGTM, have few minor comments.

scripts/src/checkprcontent/checkpr.py
mmulholla: Good catch, removed.

scripts/src/report/verifier_report.py
mmulholla: Good catch, that became unnecessary as my solution evolved. Removed.

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -20,12 +20,14 @@ def get_verifier_errors_trailer():
     return "Please run the [chart-verifier](https://github.com/redhat-certification/chart-verifier) \
 and ensure all mandatory checks pass."
 
+def get_look_at_job_output_comment():
+    return f"""To see the console output with the error messages, click the "Details" \
+link next to "CI / Chart Certification" job status towards the end of this page."""
```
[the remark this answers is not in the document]
tisutisu: For now I think it is okay display the user this message. Would like explore more if it is possible to capture stderr from github actions and will create a story if found anything.

tisutisu (commented):
LGTM

PR #160, opened 2022-09-07
check report sha

mmulholla:
Added code to check the report sha is valid and error if not. 
Also added a new test to check it all works.

See: https://issues.redhat.com/browse/HELM-355


scripts/src/report/report_info.py
```diff
@@ -98,6 +120,9 @@ def get_report_chart(report_path=None,report_info_path=None):
      print("[INFO] report chart : %s" % metadata["chart"])
      return metadata["chart"]
 
+def check_report_sha(report_path=None):
+    results
```
tisutisu, line 124: I think it might be not required here

tests/functional/step_defs/HC-19_test_report_sha.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
tisutisu, line 1: I think we need to move this test in the behave path. These are pytest-bdd files, no longer in use.

mmulholla: I think these are actually files used by both bdd and behave tests and other behave tests follow this same pattern.

tests/functional/step_defs/HC-19_test_report_sha.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.fixture
+def workflow_test():
+    test_name = 'Report with report sha value'
+    test_report = 'tests/data/report.yaml'
+    workflow_test = ChartCertificationE2ETestSingle(test_name=test_name, test_report=test_report)
+    yield workflow_test
+    workflow_test.cleanup()
+
+
+@scenario('../features/HC-16_report_sha.feature', "[HC-19-001] A partner or redhat associate submits an error-free report with report sha value")
```
tisutisu, line 24: `../features/HC-16_repor_sha.feature` doesn't exist.

tests/functional/step_defs/HC-19_test_report_sha.py
tisutisu: No I don't think so, with the behave migration we are not using the `step_defs` any more, so any step definition should be under `steps`

tests/functional/behave_features/HC-19_report_sha.feature
```diff
@@ -0,0 +1,34 @@
...
+  The sha value must match the report.
+
+  Scenario Outline: [HC-19-001] A partner or redhat associate submits an error-free report with report sha value
+    Given the vendor "<vendor>" has a valid identity as "<vendor_type>"
+    And an error-free report is used in "<report_path>"
+    When the user sends a pull request with the report
+    Then the user sees the pull request is merged
+    And the index.yaml file is updated with an entry for the submitted chart
+
+    @partners @full @reportSha
```
tisutisu, line 12: `@reportSha` is used being anywhere?

tisutisu (commented):
Overall LGTM, except that one extra step_defs file we are adding.

tests/data/HC-19/report_edited_sha_bad/report.yaml
```diff
@@ -0,0 +1,100 @@
+apiversion: v1
+kind: verify-report
+metadata:
+    tool:
+        verifier-version: 1.9.0
+        reportDigest:  uint64:17691208852129424495
+        profile:
+            VendorType: community
```
tisutisu, line 8: This chart is from `partners`, how come the VendorType is `community` here?

mmulholla: I think that was the edit I did to cause the sha to be wrong but I put it back and changed something else.

tests/functional/step_defs/HC-19_test_report_sha.py
mmulholla: Removed it

tisutisu (approved):
LGTM now. Ran the smoke tests locally, result is passing.

PR #163, opened 2022-09-29
Bump Mako==1.2.3: dependabot alert

tisutisu:
This PR is to address the dependabot alert here https://github.com/openshift-helm-charts/charts/security/dependabot/1


mmulholla (approved):
LGTM

PR #167, opened 2022-10-18
Chart Downloads v1.0 

Kartikey-star:
This PR makes changes to Chart downloads metrics made on the segment, so as to write data for top 5 charts and rank the charts with respect to number of downloads.
https://issues.redhat.com/browse/HELM-376 
Signed-off-by: Kartikey Mamgain [mamgainkartikey@gmail.com](mailto:mamgainkartikey@gmail.com)

### DIFF
--- a/scripts/src/metrics/metrics.py
+++ b/scripts/src/metrics/metrics.py
@@ -12,10 +12,11 @@ from github import Github
 sys.path.append('../')
 from indexfile import index
 from pullrequest import prepare_pr_comment as pr_comment
+from collections import OrderedDict
 
 file_pattern = re.compile(r"charts/([\w-]+)/([\w-]+)/([\w\.-]+)/([\w\.-]+)/.*")
-ignore_users=["zonggen","mmulholla","dperaza4dustbit","openshift-helm-charts-bot","baijum","tisutisu","rhrivero"]
-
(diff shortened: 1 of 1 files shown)


scripts/src/metrics/metrics.py
```diff
@@ -57,7 +58,29 @@ def send_release_metrics(write_key, downloads):
...
-            send_metric(write_key,provider,f"{chart} downloads", metrics[provider][chart])
+            number_of_downloads=0
+            for key,value in metrics[provider][chart].items():
+                chart_downloads.append({"downloads":number_of_downloads,"name":key,"provider":provider})
+    chart_downloads.sort(key = lambda i: i[0],reverse=True)
+
+    for x in range(len(chart_downloads)):
+        send_download_metric(write_key,chart_downloads[x]["provider"],chart_downloads[x]["downloads"],chart_downloads[x]["name"],x+1)
+
+    for x in range(5):
```
mmulholla, line 69: One thing here we don't want the more than one version of any chart in the top 5. Also only the latest version of the chart should be considered. So if a chart has version 1.0.0 at 500 downloads and version 1.1.0 at 200 downloads we  would only consider 1.1.0 with 200 downloads for the top 5.

Kartikey-star: Only for the top 5 , different versions should not be considered right, but they still need to written insegment with the different id

mmulholla (approved):
LGTM

PR #170, opened 2022-10-26
Owner Metrics

Kartikey-star:
This PR aims at creating a new workflow to be triggered on a push request. If the request is from the redhat-mercury-bot for an OWNERS file create or update a new stat should be listed in the segment.
Please refer to story https://issues.redhat.com/browse/HELM-417 for more details.

### DIFF
--- a/scripts/setup.cfg
+++ b/scripts/setup.cfg
@@ -48,4 +48,5 @@ console_scripts =
     check-user = owners.checkuser:main
     metrics = metrics.metrics:main
     get-verify-params = report.get_verify_params:main
+    pushowners=metrics.pushowners:main
 
--- a/scripts/src/owners/owners_file.py
+++ b/scripts/src/owners/owners_file.py
@@ -8,20 +8,53 @@ except ImportError:
 
 
 def get_owner_data(category, organization, chart):
+    path=os.path.join("charts", category, organization, chart, "OWNERS")
+    status,owner_content=get_owner_data_from_file(path)
+    return status,owner_content
+
+def get_owner_data_from_file(owner_path):
     try:
-        with open(os.path.join("charts", category, organization, chart, "OWNERS")) as owner_data:
+        with open(owner_path) as owner_data:
             owner_content = yaml.load(owner_data,Loader=Loader)
         return True,owner_content
     except Exception as err:
         print(f"Exception loading OWNERS file: {err}")
         return False,""
 
+def get_vendor(owner_data):
+    vendor=""
+    try:
+        vendor = owner_data['vendor']['name']
+    except Exception:
+        pass
+    return vendor
+
+def get_chart(owner_data):
+    chart=""
+    try:
+        chart = owner_data['chart']['name']
+    except Exception:
+        pass
+    return chart
+
 
 def get_provider_delivery(owner_data):
     provider_delivery = False
     try:
-        provider_delivery = owner_data["providerDelivery"]
+        provider_delivery = owner_data['providerDelivery']
     except Exception:
         pass
     return provider_delivery
+
+def get_users_included(owner_data):
+    users_included="No"
+    try:
+        users = owner_data['users']
+        if len(users)!=0:
+            return "Yes"
+    except Exception:
+        pass
+    return users_included
+
+
 
--- a/.github/workflows/owners.yml
+++ b/.github/workflows/owners.yml
@@ -1,0 +1,86 @@
+name: Metrics for owners file
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  owners-metrics:
+    name: Send Owner Metrics
+    runs-on: ubuntu-20.04
+    env:
+      SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY }}
+      SEGMENT_TEST_WRITE_KEY: ${{ secrets.SEGMENT_TEST_WRITE_KEY }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Exit if push is not from redhat-mercury-bot
+        id: check_push_by_bot
+        run: |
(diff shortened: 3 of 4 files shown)


scripts/src/pushowners/owners.py
```diff
@@ -0,0 +1,109 @@
...
+    print(f"   --modified : {args.modified}")
+    print(f"   --repository : {args.repository}")
+    print(f"   --prefix : {args.prefix}")
+
+    if not args.write_key:
+        print("Error: Segment write key not set")
+        sys.exit(1)
+
+    users_included,provider_delivery,vendor_name,chart_name,vendor_type,action,update = process_pr(args.added,args.modified)
+    send_owner_metric(args.write_key,args.prefix,users_included,provider_delivery,vendor_name,chart_name,vendor_type,action,update)
```
mmulholla, line 106: Code appears to assume the file changed is an OWNERS file but we should check that. Could even do that in the workflow.

scripts/src/pushowners/owners.py
```diff
@@ -0,0 +1,109 @@
...
+
+def getFileContent(files):
+    users_included="No"
+    provider_delivery="No"
+    vendor_name=""
+    chart_name=""
+    vendor_type=""
+    for changed_file in files:
+        # Load the YAML file
+        with open(changed_file) as file:
```
mmulholla, line 17: We should require the file name to be charts/partners/<vendor>/OWNERS - we are not really interested in redhat or community charts because OWNERs file for these are not done by the redhat-mercury-bot. Also code would try to get data from the file even if it was not an OWNERS file.

Also the loop allows for multiple files but the function only returns data for one - the first found. I would suggest either we allow a PR to contain multiple OWNERS files and we do a metric for each or only get a metric if there is only one file in the push request and it is an  OWNERS file. Currently I would only expect the bot to include one OWNERS file so that would be my choice.

scripts/setup.cfg
```diff
@@ -48,4 +48,5 @@ console_scripts =
     check-user = owners.checkuser:main
     metrics = metrics.metrics:main
     get-verify-params = report.get_verify_params:main
+    pushowners=pushowners.owners:main
```
mmulholla, line 51: I think this  new python scripts should be under the script/metrics directory. We have too many directories already and the names gets confusing.

scripts/src/pushowners/owners.py
```diff
@@ -0,0 +1,109 @@
...
+            documents = yaml.full_load(file)
+            for key, value in documents.items():
+                if key=='providerDelivery' and value=='True':
+                    provider_delivery="Yes"
+                elif key=='users' and len(key)!=0:
+                    users_included="Yes"
+                elif key=='vendor':
+                    vendor_name=value['name']
+                elif key=='chart':
+                    chart_name=value['name']
```
mmulholla, line 27: I would like this code to make use of,  and be in,  scripts/owners/owners_file.py which is meant to be a one stop shop for all things connected to OWNERS files. It loads the files etc, so add some methods there which return the fields you want. This way if the OWNERS file format ever changes (e.g: the key names change) the only thing we need to change is owners_file.py.

mmulholla (changes requested):
Have added some comments.

scripts/setup.cfg
Kartikey-star: So as per the below comments we need to move the scripts to scripts/owners directory.

Kartikey-star: Sure, makes sense

scripts/src/pushowners/owners.py
Kartikey-star: Yes will update that

scripts/src/pushowners/owners.py
Kartikey-star: Makes sense

scripts/src/pushowners/owners.py
Kartikey-star: I was actually having a loop as the modified and added files returned by jitterbit/get-changed-files@v1 is a list. 
But yeah as we are aware that the bot would push just one file , i have made changes to the code to accept a single file.

scripts/src/pushowners/owners.py
Kartikey-star: Made changes to worflow to validate the same.

scripts/src/owners/owners_file.py
```diff
@@ -25,3 +25,26 @@ def get_provider_delivery(owner_data):
         pass
     return provider_delivery
 
+def getFileContent(changed_file):
```
mmulholla, line 28: This is not quit how I envisaged this.
- the code loads the yaml in two places now, make a load from file method and have two existing cases call it.
- I would then expect a separate method to get each attribute. Calling code then just use the get method which return null if it does exist.  
- Getting the vendor type if not an owners file function so would not expect that part in this file but as part of the caller. 

For info I was extending the owners_file.py for the signed chart work and finished with this which you may use:
```
def get_owner_data(category, organization, chart):
    return get_owner_data_from_file(os.path.join("charts", category, organization, chart, "OWNERS"))

def get_owner_data_from_file(owner_path):
    try:
        with open(owner_path) as owner_data:
            owner_content = yaml.load(owner_data,Loader=Loader)
        return True,owner_content
    except Exception as err:
        print(f"Exception loading OWNERS file: {err}")
        return False,""
```

mmulholla (changes requested):

scripts/src/owners/owners_file.py
Kartikey-star: the reason i did not use the get_owner_data function in owners_file.py is because we are getting the file paths from the action itself. So i do not need to pass in category,organization and chart as params to the get_owner_data and hence added the entire method. Let me know if it is fine.

mmulholla (approved):
LGTM

PR #173, opened 2022-11-02
Catching stderr in case command fails

tisutisu:
This PR is to fix the HC-19 test case failure


tisutisu:
Closing this as the issue is on the chart-verifier side.

PR #175, opened 2022-11-08
Fix tests - report sha message case

mmulholla:
Problem was cause by a change to the chart verifier where a message was changed from:
- Digest in report did not match report content 
to:
- digest in report did not match report content

so Digest -> digest

Unfortunately this broke the test,


tisutisu (approved):
LGTM

PR #186, opened 2022-11-21
Updates git commit log with indexfile name

tisutisu:
when providerDelivery flag was set to true, in the commit history log will print the exact file it changed.
Observed wrong commit messge issue here https://github.com/openshift-helm-charts/stage/pull/16


mmulholla (approved):
LGTM

PR #192, opened 2022-12-12
Update provider delivery message to web catalog only

PatAKnight:
Jira: https://issues.redhat.com/browse/HELM-462

Updates the error messages whenever there is an issue with provider delivery to now display web catalog only. 

The mention of provider controlled delivery in the error messages can lead to some confusions with users. This is due to the fact that the [Creating a Helm Chart Certification Project](https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/helm-chart-certification/creating-a-helm-chart-certification-project) article does not have any mentions of a provider delivery distribution method. In our case provider delivery is the same as the web catalog only distribution method.

### DIFF
--- a/scripts/src/checkprcontent/checkpr.py
+++ b/scripts/src/checkprcontent/checkpr.py
@@ -54,7 +54,7 @@ def check_provider_delivery(report_in_pr,num_files_in_pr,report_file_match):
     provider_delivery = False
     if report_in_pr and num_files_in_pr > 1:
         if report_provider_delivery or owner_provider_delivery:
-            msg = f"[ERROR] OWNERS file and/or report indicate web catalog only delivery but pull request is not report only. The distribution method web catalog only requires providerDelivery to be set to true in the OWNERS file."
+            msg = f"[ERROR] The web catalog distribution method requires the pull request to be report only."
             print(msg)
             print(f"::set-output name=pr-content-error-message::{msg}")
             sys.exit(1)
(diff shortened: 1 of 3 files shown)


scripts/src/checkprcontent/checkpr.py
```diff
@@ -54,7 +54,7 @@ def check_provider_delivery(report_in_pr,num_files_in_pr,report_file_match):
     provider_delivery = False
     if report_in_pr and num_files_in_pr > 1:
         if report_provider_delivery or owner_provider_delivery:
-            msg = f"[ERROR] OWNERS file and/or report indicate provider controlled delivery but pull request is not report only."
+            msg = f"[ERROR] OWNERS file and/or report indicate web catalog only delivery but pull request is not report only. The distribution method web catalog only requires providerDelivery to be set to true in the OWNERS file."
```
mmulholla, line 57: I wanted to get away from referring the "OWNERS file" and instead "the distribution method selected"

mmulholla (commented):
small comment on message content.

scripts/src/checkprcontent/checkpr.py
PatAKnight: So something like "Web catalog only requires providerDelivery to be set to true in the distribution method selected"?

mmulholla (approved):
LGTM

PR #201, opened 2023-01-06
Web catalog only

mmulholla:
add code so that provider delivery and web catalog only are treated as he same thing.  

also fix failing builds:
add authorization token to header for github api requests
correctly process response when request return a failure message.
Moved code o get files and label to a common function which caches results.
output api rate limit data after api calls.

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -22,6 +22,7 @@ sys.path.append('../')
 from report import report_info
 from report import verifier_report
 from signedchart import signedchart
+from pullrequest import prartifact
 
 def write_error_log(directory, *msg):
     os.makedirs(directory, exist_ok=True)
@@ -39,20 +40,12 @@ def get_vendor_type(directory):
         sys.exit(1)
     return vendor_type
 
-def get_labels(api_url):
-    # api_url https://api.github.com/repos/<organization-name>/<repository-name>/pulls/1
-    headers = {'Accept': 'application/vnd.github.v3+json'}
-    r = requests.get(api_url, headers=headers)
-    return r.json()["labels"]
-
 def get_modified_charts(directory, api_url):
     print("[INFO] Get modified charts. %s" %directory)
-    files_api_url = f'{api_url}/files'
-    headers = {'Accept': 'application/vnd.github.v3+json'}
-    r = requests.get(files_api_url, headers=headers)
+    files = prartifact.get_modified_files(api_url)
     pattern = re.compile(r"charts/(\w+)/([\w-]+)/([\w-]+)/([\w\.-]+)/.*")
-    for f in r.json():
-        m = pattern.match(f["filename"])
+    for file_path in files:
+        m = pattern.match(file_path)
         if m:
             category, organization, chart, version = m.groups()
             return category, organization, chart, version
@@ -260,8 +253,7 @@ def check_report_success(directory, api_url, report_path, report_info_path, vers
 
     report = report_info.get_report_results(report_path=report_path,report_info_path=report_info_path,profile_type=vendor_type)
 
-    labels = get_labels(api_url)
-    label_names = [l["name"] for l in labels]
+    label_names = prartifact.get_labels(api_url)
 
     failed = report["failed"]
     passed = report["passed"]
@@ -367,7 +359,7 @@ def main():
     generated_report_path = os.environ.get("GENERATED_REPORT_PATH")
     generated_report_info_path =  os.environ.get("REPORT_SUMMARY_PATH")
     env = Env()
-    provider_delivery = env.bool("PROVIDER_DELIVERY",False)
+    web_catalog_only = env.bool("WEB_CATALOG_ONLY",False)
 
     if os.path.exists(submitted_report_path):
         print("[INFO] Report exists: ", submitted_report_path)
@@ -376,7 +368,7 @@ def main():
         report_info_path = ""
         if report_generated and report_generated == "True":
             match_checksum(args.directory,generated_report_info_path, category, organization, chart, version)
-        elif not provider_delivery:
+        elif not web_catalog_only:
             check_url(args.directory, report_path)
     else:
         print("[INFO] Report does not exist: ", submitted_report_path)
--- a/scripts/src/chartrepomanager/chartrepomanager.py
+++ b/scripts/src/chartrepomanager/chartrepomanager.py
@@ -197,6 +197,7 @@ def set_package_digest(chart_entry):
 
     url = chart_entry["urls"][0]
     head = requests.head(url, allow_redirects=True)
+
     target_digest = ""
     if head.status_code == 200:
         response = requests.get(url, allow_redirects=True)
--- a/scripts/src/chartrepomanager/chartrepomanager.py
+++ b/scripts/src/chartrepomanager/chartrepomanager.py
@@ -2,6 +2,7 @@ import argparse
 import shutil
 import os
 import sys
+import json
 import re
 import subprocess
 import tempfile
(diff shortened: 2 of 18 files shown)


scripts/src/pullrequest/prartifact.py
```diff
@@ -10,33 +10,79 @@
...
+        r = requests.get(api_url, headers=headers)
+        pr_data = r.json()
+
+        if xRateLimit in r.headers:
+            print(f'[DEBUG] {xRateLimit} : {r.headers[xRateLimit]}')
+        if xRateRemain in r.headers:
+            print(f'[DEBUG] {xRateRemain}  : {r.headers[xRateRemain]}')
+
+        if "message" in pr_data:
+            print(f'[ERROR] getting pr files: {labels["message"]}')
```
tisutisu, line 78: I think you meant `pr_data["message"]` here instead of `labels["message"]`

scripts/src/pullrequest/prartifact.py
```diff
@@ -10,33 +10,79 @@
...
+                print(f'[DEBUG] {xRateRemain}  : {r.headers[xRateRemain]}')
+
+            if "message" in files:
+                print(f'[ERROR] getting pr files: {files["message"]}')
+                sys.exit(1)
+            else:
+                for file in files:
+                    if "filename" in file:
+                        pr_files.append(file["filename"])
+                    if "message" in file:
```
tisutisu, line 58: I think this condition check is dead code here, since we are already checking earlier if `"message" in files` and exiting if it is true.

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -218,12 +218,13 @@ def set_package_digest(chart_entry):
         raise Exception("Was unable to compute SHA256 digest, please ensure chart url points to a chart package.")
 
 
-def update_index_and_push(indexfile,indexdir, repository, branch, category, organization, chart, version, chart_url, chart_entry, pr_number, provider_delivery):
+def update_index_and_push(indexfile, indexdir, repository, branch, category, organization, chart, version, chart_url, chart_entry, pr_number, web_catalog_only):
     token = os.environ.get("GITHUB_TOKEN")
     print(f"Downloading {indexfile}")
     r = requests.get(f'https://raw.githubusercontent.com/{repository}/{branch}/{indexfile}')
```
tisutisu, line 224: Here also we making https request, I think we should include BOT_TOKEN here as well.
I found BOT_TOKEN missing while calling `requests.get` in this module above under function `set_package_digest(chart_entry)`

scripts/src/checkautomerge/checkautomerge.py
```diff
@@ -1,18 +1,25 @@
...
     # api_url https://api.github.com/repos/<organization-name>/<repository-name>/pulls/1
-    headers = {'Accept': 'application/vnd.github.v3+json'}
+    headers = {'Accept': 'application/vnd.github.v3+json','Authorization': f'Bearer {os.environ.get("BOT_TOKEN")}'}
     merged = False
     for i in range(20):
         r = requests.get(api_url, headers=headers)
-        if r.json()["merged"]:
+        response_content = r.json()
+        if "message" in response_content:
+            print(f'[ERROR] getting public key: {response_content["message"]}')
```
tisutisu, line 16: I think you meant `[Error] getting merge status` here

scripts/src/saforcertadmin/push_secrets.py
```diff
@@ -46,16 +50,29 @@ def get_repo_public_key(repo):
...
+
     return response_json['key_id'], response_json['key']
 
 def get_repo_secrets(repo):
     """Get the list of secret names of a github repository"""
     secret_names = []
     response = requests.get(f'https://api.github.com/repos/{repo}/actions/secrets', headers=headers)
     if response.status_code != 200:
-        logging.error(f"unexpected response getting secrets : {response.status_code} : {response.reason}")
+        logging.error(f"[ERROR] unexpected response getting public key : {response.status_code} : {response.reason}")
```
tisutisu, line 70: I think error message was correct earlier.. no need to change here.

scripts/src/saforcertadmin/push_secrets.py
```diff
@@ -46,16 +50,29 @@ def get_repo_public_key(repo):
...
         sys.exit(1)
     response_json = response.json()
+    if "message" in response_json:
+        print(f'[ERROR] getting public key: {response_json["message"]}')
```
tisutisu, line 74: We can improve the error message here .. I think `[ERROR] getting repo secret : ` would sound better

scripts/src/saforcertadmin/push_secrets.py
```diff
@@ -66,7 +83,15 @@ def create_or_update_repo_secrets(repo, secret_name, key_id, encrypted_value):
     if response.status_code != 201 and response.status_code != 204:
         logging.error(f"unexpected response during put request : {response.status_code} : {response.reason}")
         sys.exit(1)
-    #response_json = response.json()
+    try:
+        response_json = response.json()
+        if "message" in response_json:
+            print(f'[ERROR] updating public key: {response_json["message"]}')
```
tisutisu, line 89: Also improve the error message here as `[ERROR] updating repo secrets: `

scripts/src/tools/gitutils.py
```diff
@@ -44,10 +45,26 @@ def set_git_username_email(repo, username, email):
...
+        pass
+
+
     return r
 
 def github_api_get(endpoint, bot_token, headers={}):
     r = requests.get(f'{GITHUB_BASE_URL}/{endpoint}', headers=headers)
+    response_json = r.json()
+    if "message" in response_json:
+        print(f'[ERROR] getting secret: {response_json["message"]}')
```
tisutisu, line 65: We can improve the error message saying `[ERROR] During GET Request :`

tisutisu (changes requested):
Changes are looking good, except few minor logging nits. Please take a look in the comments. Also would like to see the BOT_TOKEN being added in two more places https://github.com/openshift-helm-charts/development/blob/main/scripts/src/metrics/metrics.py#L447 and  https://github.com/openshift-helm-charts/development/blob/main/scripts/src/metrics/metrics.py#L39 since we might hit the API limit in these places as well, just to be in the safe side.

scripts/src/pullrequest/prartifact.py
mmulholla: Yes, thx, fixed it

scripts/src/pullrequest/prartifact.py
mmulholla: Good spot, removed it

scripts/src/chartprreview/chartprreview.py
```diff
@@ -376,7 +368,7 @@ def main():
         report_info_path = ""
         if report_generated and report_generated == "True":
             match_checksum(args.directory,generated_report_info_path, category, organization, chart, version)
-        elif not provider_delivery:
+        elif not web_catalog_only:
             check_url(args.directory, report_path)
     else:
         print("[INFO] Report does not exist: ", submitted_report_path)
```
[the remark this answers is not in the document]
mmulholla: The request in verify-package digest is not a github api call. just a request to get the file in the chart url. I double checked to get the api call before and after and not change.

scripts/src/chartrepomanager/chartrepomanager.py
mmulholla: Similar to above the call is not a github api call, call to a chart url.

scripts/src/checkautomerge/checkautomerge.py
mmulholla: Thx, updated.

scripts/src/saforcertadmin/push_secrets.py
mmulholla: Yes was trying to see if codeql was complaining about the word "secret" and forgot to change it back.

scripts/src/tools/gitutils.py
mmulholla: good spot, fixed it.

scripts/src/saforcertadmin/push_secrets.py
mmulholla: as above, updated.

scripts/src/saforcertadmin/push_secrets.py
mmulholla: as above again, updated.

tisutisu (approved):
LGTM

PR #203, opened 2023-01-11
Fixed the metrics workflow failure

tisutisu:
This PR will fix the nightly metrics workflow run failure here https://github.com/openshift-helm-charts/development/actions/runs/3888745805/jobs/6636375690

### DIFF
--- a/scripts/src/metrics/metrics.py
+++ b/scripts/src/metrics/metrics.py
@@ -47,10 +47,10 @@ def get_release_metrics():
             sys.exit(1)
             
         response_json = response.json()
-        if xRateLimit in response_json.headers:
-           print(f'[DEBUG] {xRateLimit} : {response_json.headers[xRateLimit]}')
-        if xRateRemain in response_json.headers:
-            print(f'[DEBUG] {xRateRemain}  : {response_json.headers[xRateRemain]}')
+        if xRateLimit in response.headers:
+           print(f'[DEBUG] {xRateLimit} : {response.headers[xRateLimit]}')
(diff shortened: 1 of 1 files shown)


scripts/src/metrics/metrics.py
```diff
@@ -47,10 +47,10 @@ def get_release_metrics():
...
             
         response_json = response.json()
-        if xRateLimit in response_json.headers:
-           print(f'[DEBUG] {xRateLimit} : {response_json.headers[xRateLimit]}')
-        if xRateRemain in response_json.headers:
-            print(f'[DEBUG] {xRateRemain}  : {response_json.headers[xRateRemain]}')
+        if xRateLimit in response_json:
+           print(f'[DEBUG] {xRateLimit} : {response_json[xRateLimit]}')
+        if xRateRemain in response_json:
+            print(f'[DEBUG] {xRateRemain}  : {response_json[xRateRemain]}')
```
mmulholla, line 53: I don't think this is the correct fix. The mistake in the original was using ```reponse_json.headers``` - it should of been using ```reponse.headers```. New code will not error but will also never produce any output.

mmulholla (changes requested):

scripts/src/metrics/metrics.py
tisutisu: Changed `response_json[xRateLimit]` to response.headers[xRateLimit]` in the latest commit.

mmulholla (approved):
LGTM

PR #208, opened 2023-01-20
Fix: Always send the notification evenif the previous step fails

tisutisu:
This will fix the notification not being sent when the previous step fails.


mmulholla (approved):
LGTM

PR #212, opened 2023-02-02
Added sleep for the intermittent issue

tisutisu:
Added sleep of 10 seconds before update_index_and_push function.


mmulholla (commented):
LGTM to see if it helps.

mmulholla (approved):

PR #225, opened 2023-05-11
Fix variable reference before assignment in generic fail comment

komish:
The generic failure did not initialize the `msg` variable, so writing the PR content would fail.


jsm84 (approved):
LGTM

PR #226, opened 2023-05-15
adjust case of an error prefix produced by chart-verifier

komish:
Chart Verifier recently ran style tooling which enforced that error messages should not start with capital letters, or end with punctuation. As a result, this feature contains a message string that directly corresponded to a capital letter that was previously used. This PR replaces that with a lower case letter to align the two.


jsm84 (approved):
LGTM

PR #232, opened 2023-06-15
Ensure chart annotations are merged with generated annotations on index creation

komish:
Fixes: https://github.com/openshift-helm-charts/charts/issues/894

When chart-only PRs are submitted, our automation extracts annotations that would otherwise have been submitted in a report from the report generated on the partner's behalf. These annotations were then replacing the existing annotations in a given chart instead of merging with them.

This PR merges the annotations with existing annotations in the chart, replacing overlapping keys with the values that were generated by our automation.


jsm84 (approved):
LGTM

PR #242, opened 2023-07-26
Update signed charts based on new fixtures

komish:
One more fixture update here that I can see. Needed to have the first group merge so I could target them remotely. Now I need to target this one.


jsm84 (approved):
LGTM

PR #243, opened 2023-07-26
Regenerate test reports after chart modifications

komish:
In the recent move to OCP 4.13 underpinning this repo's tests, one of the test charts would not run. That's been resolved, but in resolving that, we needed to regenerate various permutations of reports that use those charts as a base. 

This PR regenerates those reports.


komish:
It's unclear to me whether I'm expecting the tests to pass, but I'l' reevaluate that once I see the outcome of the run.

jsm84 (approved):

PR #247, opened 2023-08-07
Apply black and fix some PEP8 violations

mgoerens:
This fixes only some PEP8 issues as some fixes would require further testing.

Here I focused on easy fixes with a low risk (unused imports, f-string without placeholders, ...), in addition to reformatting with black.


komish (approved):
/lgtm

PR #253, opened 2023-09-11
Updated reports for vault chart fixtures

jsm84:
Unblocks #250

Summary:
* Regenerated reports for the vault charts that were updated due to recent breakage from a chart-verifier bugfix.


komish (approved):
/lgtm

PR #255, opened 2023-09-12
Add comments to part of the python scripts

mgoerens:


scripts/src/chartprreview/chartprreview.py
```diff
@@ -413,17 +501,24 @@ def main():
...
+        args.directory, category, organization, chart
     )
+
+    report_generated = os.environ.get("REPORT_GENERATED")
+    generated_report_path = os.environ.get("GENERATED_REPORT_PATH")
+    generated_report_info_path = os.environ.get("REPORT_SUMMARY_PATH")
+    env = Env()
+    web_catalog_only = env.bool("WEB_CATALOG_ONLY", False)
+
     submitted_report_path = os.path.join(
```
komish, line 519: Can you elaborate a bit on what's happening here and why? I see some things got moved up, but I also see some params are being changed? These are more than documentation/comment changes, so I want to be sure of the intention behind them. My guess is unused parameters.

scripts/src/owners/owners_file.py
```diff
@@ -66,9 +66,4 @@ def get_users_included(owner_data):
 
 
 def get_pgp_public_key(owner_data):
-    pgp_public_key = "null"
-    try:
-        pgp_public_key = owner_data["publicPgpKey"]
-    except Exception:
-        pass
-    return pgp_public_key
+    return owner_data.get("publicPgpKey", "")
```
komish, line 69: Have you checked the logic where this is called, and how it handles `""` vs. `"null"`? I know that we stuff "null" in some places, though I'm not certain as to why. To maintain compatibility here without diving deeper, I'd probably switch the default value if the key isn't found in that map to "null".


```suggestion
    return owner_data.get("publicPgpKey", "null")
```

scripts/src/report/verifier_report.py
```diff
@@ -171,7 +223,7 @@ def validate(report_path):
             v1_0_profile = False
             if profile_version.major == 1 and profile_version.minor == 0:
                 v1_0_profile = True
-        except Exception:
+        except ValueError:
```
komish, line 226: 1. Do we want to ignore other exceptions types?
2. Are we certain the Version.coerce call that's likely to through an error here is going to throw a ValueError?

scripts/src/chartprreview/chartprreview.py
```diff
@@ -105,7 +138,15 @@ def check_owners_file_against_directory_structure(
         sys.exit(1)
 
 
-def verify_signature(directory, category, organization, chart, version):
+def verify_signature(category, organization, chart, version):
```
komish, line 141: Removing directory here seems fine, but I'm wondering: should we have instead leveraged the directory to return an error?

I'm thinking that typically this directory is where PR "artifacts" live... that is, things that typically make it back to the user in the PR as a comment indicating what went wrong.

We covered that directory here is likely unused, but should we have used that directory to communicate some error in this block back to the user instead of removing it?

mgoerens: I understand what you mean.

Currently the way errors are handled (at least in the rest of this file) is by writing them to the logs, then exiting with a return code of 1:
```
            write_error_log(directory, msg)
            sys.exit(1)
```

I'm not a fan of this approach, because it means having `sys.exit` deep inside multiple function calls. At some point we should probably review it. For instance, when we harmonize the way github comments are written across the different certification projects.

Also, at this moment I don't see where the artifact directory is used. Besides writing error logs, it seems to be adding some other file metadata (in `prartifact.py`). But is it made eventually available to the user ? As a directory to download ? I can't find this.

So, back to your initial comment, regarding this function. Currently it never fails AFAICT. I can add some error logs and `sys.exit` if the check fails.

komish: > I'm not a fan of this approach, because it means having sys.exit deep inside multiple function calls. At some point we should probably review it.

100%. Couldn't agree more.

With this in mind, we can re-evaluate whether or not we need to re-add the directory here in the future when we normalize error messages. I'll mark this topic as resolved.

komish (approved):
/lgtm

PR #256, opened 2023-09-13
Adjust expected timeout message and accept regex in expected message

mgoerens:
This PR **should**  fix the pipeline for https://github.com/openshift-helm-charts/development/pull/254

The reason for the failure is because in scenario `HC-16_chart_test_takes_more_than_30mins` for partners, we're expecting the string `* timed out waiting for the condition` in the PR comment. The time out does occur, and the PR does get a comment about a timeout, but it doesn't contains this exact string.
This is the PR in question FYI: https://github.com/openshift-helm-charts/sandbox/pull/65858

I dug around but I can't seem to understand how this was passing until now. They were a few recent changes that could have affected this, but I must be missing something because I can't find the culprit:
* github.com/helm/chart-testing/v3 updated from v3.8.0 to v3.9.0 in chart-verifier
* https://github.com/redhat-certification/chart-verifier/pull/405
* https://github.com/openshift-helm-charts/development/pull/235

### DIFF
--- a/tests/functional/behave_features/HC-16_chart_test_takes_more_than_30mins.feature
+++ b/tests/functional/behave_features/HC-16_chart_test_takes_more_than_30mins.feature
@@ -10,8 +10,8 @@ Feature: Chart test takes longer time and exceeds default timeout
 
     @partners @full
     Examples:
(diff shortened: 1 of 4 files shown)


komish (approved):
/lgtm

PR #261, opened 2023-09-15
Update vault-timeout chart to fix timeout errors

jsm84:
Summary:
* Yes, you read that right. HC-16 behave test (which tests timeout) is exiting abnormally with a timeout error, when the timeout message should only show up in the report.


komish (approved):
/lgtm

PR #264, opened 2023-09-22
Get OCP range from external library

mgoerens:


.github/actions/get-ocp-range/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+          go-version: '>=1.20'
+
+    - name: Install get-ocp-range
+      shell: bash
+      run: go install github.com/opdev/getocprange@latest
+
+    - name: Run get-ocp-range
+      id: run-get-ocp-range
+      shell: bash
+      run: echo "ocp-version-range=$(get-ocp-range '${{ inputs.kube-version-range }}')" >> $GITHUB_OUTPUT
```
komish, line 26: If `get-ocp-range` throw an error, does this fail? 

If not, lets add steps to make sure we got valid output from `get-ocp-range`. 
Bonus points if you [set an error message](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message) if it fails, or a debug (or other verbosity) message if it succeeds, with the value `get-ocp-range` returned.

.github/workflows/build.yml
```diff
@@ -233,9 +246,13 @@ jobs:
...
           REPORT_SUMMARY_PATH: ${{ steps.run-verifier.outputs.report_info_file }}
           WORKFLOW_WORKING_DIRECTORY: "../pr"
+          OCP_VERSION_RANGE: ${{ steps.get-ocp-range.outputs.ocp-version-range }}
         run: |
           cd pr-branch
-          ../ve1/bin/chart-pr-review --directory=../pr --verify-user=${{ github.event.pull_request.user.login }} --api-url=${{ github.event.pull_request._links.self.href }}
+          ../ve1/bin/chart-pr-review \\
+            --directory=../pr \\
+            --verify-user=${{ github.event.pull_request.user.login }} \\
+            --api-url=${{ github.event.pull_request._links.self.href }}
```
komish, line 255: Are double `\\` needed here? Shouldn't this be single `\`?

scripts/src/indexfile/index.py
```diff
@@ -111,7 +111,7 @@ def get_latest_charts():
 
     return latest_charts
 
-
+# TODO: never called ? is this just a test ??
 if __name__ == "__main__":
     get_chart_info("redhat-dotnet-0.0.1")
```
komish, line 116: Instead of removing the comment, should we remove the block?

scripts/src/report/verifier_report.py
```diff
@@ -204,27 +204,33 @@ def validate(report_path):
...
+            #         else:
+            #             return (
+            #                 False,
+            #                 f"Kube Version {chart[KUBE_VERSION_ATTRIBUTE]} translates to an invalid OCP version range {kube_supported_ocp_versions_string}",
+            #             )
+            # else:
+            #     if v1_0_profile:
+            #         return True, ""
+            #     else:
+            #         return False, f"{KUBE_VERSION_ATTRIBUTE} missing from chart!"
```
komish, line 233: Think we can remove this instead of commenting?

komish (changes requested):
Couple of small things noted.

scripts/src/report/verifier_report.py
mgoerens: Are v1.0 profiles still supported ?

This piece of code contains logic to ignore errors when the Kube Version Range is not set in a report (for profile v1.0 only). Should I reimplement this logic ?

FYI currently I counted 30 reports that are using the v1.0 profile:
```
❯ for i in $(find charts/ -name report.yaml); do profile_version=$(podman run --rm -v "${PWD}":/workdir:z mikefarah/yq .metadata.tool.profile.version $i); if grep -q "1.0" <<< $profile_version; then echo $i $profile_version; fi; done
charts/community/redhat/dotnet/0.0.1/report.yaml v1.0
charts/community/redhat/wildfly/1.4.0/report.yaml v1.0
charts/community/redhat/wildfly/1.5.2/report.yaml v1.0
charts/partners/fortanix/dsm-secrets-injector-chart/2.0/report.yaml v1.0
charts/partners/fortanix/dsm-secrets-injector-chart/2.2-4.6/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.10/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.12/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.13/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.14/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.3/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.4/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.5/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.6/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.7/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.8/report.yaml v1.0
charts/partners/ibm/ibm-object-storage-plugin/2.1.9/report.yaml v1.0
charts/partners/ibm/ibm-oms-ent-prod/6.0.2/report.yaml v1.0
charts/partners/ibm/ibm-oms-pro-prod/6.0.2/report.yaml v1.0
charts/partners/ibm/ibm-operator-catalog-enablement/1.2.2/report.yaml v1.0
charts/partners/ibm/ibm-spectrum-protect-plus-prod/1.2.1/report.yaml v1.0
charts/partners/nirmata/kyverno/v2.1.2/report.yaml v1.0
charts/partners/test-org/psql-service/0.1.10/report.yaml v1.0
charts/redhat/redhat/dotnet/0.0.1/report.yaml v1.0
charts/redhat/redhat/eap-xp3/1.0.0/report.yaml v1.0
charts/redhat/redhat/eap74/1.1.0/report.yaml v1.0
charts/redhat/redhat/nodejs-ex-k/0.2.1/report.yaml v1.0
charts/redhat/redhat/nodejs/0.0.1/report.yaml v1.0
charts/redhat/redhat/psql-sed/1.0.0/report.yaml v1.0
charts/redhat/redhat/quarkus/0.0.3/report.yaml v1.0
charts/redhat/redhat/spring-boot-rest-http/0.0.1/report.yaml v1.0
```

mgoerens: The issue is that (with my changes) the pipeline is likely to fail for reports with profile = 1.0, as the step which runs get-ocp-range will fail.

I believe I need to extract the profile version as part of the pipeline and handle the 1.0 case gracefully there (e.g. set a default value for the OCP_RANGE.

komish: The profile versioning seems to an incremental change to expectations imposed on a given chart. To that end, I would expect all new charts to come in at the latest profile.

I'm unsure of the motivation behind allowing backwards compatibility with the profile versions. I would be interested in removing this concept of profile versioning.

`get-ocp-range` needs the kubeVersion field, right? That's been required since profile version 1.0 as I understand it. What is it that will break here for these?

mgoerens: get-ocp-range will fail if kubeVersion is missing.

Maybe the yq command will have failed also beforehand: https://github.com/openshift-helm-charts/development/pull/264/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R229

.github/actions/get-ocp-range/action.yaml
mgoerens: Going for the bonus point here

.github/actions/get-ocp-range/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+  using: "composite"
+  steps:
+    - name: Setup Go
+      uses: actions/setup-go@v4
+      with:
+          go-version: '>=1.20'
+
+    - name: Install get-ocp-range
+      shell: bash
+      run: go install github.com/opdev/getocprange@latest
```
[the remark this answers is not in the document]
mgoerens: I've changed it, but according to my testing "@latest" is still needed

scripts/src/report/verifier_report.py
mgoerens: I've implemented the following:

* added a step that extracts the profile version set in report.yaml
* added logic to fail the job if there is an error in get-ocp-range, only if the profile version is not 1.0

This is feature equivalent, as prior to this PR we were also attempting (in python) to calculate the OCP range and ignoring failures if the profile was 1.0.

On the down side it adds two extra steps to an already very long job.

> To that end, I would expect all new charts to come in at the latest profile.

If we decide to enforce this, we can simplify several parts of our codebase. I'd be in favor :+1:

komish:
@mgoerens Test Workflow still didn't seem to test the workflow here, and I think that's because of a bug.

Before we review this further, lets get https://github.com/openshift-helm-charts/development/pull/270 merged in, and this PR rebased on that to make tests run for this PR.

komish:
@mgoerens My apologies, your recent CI failure has a formatting problem that I introduced in a PR before the Black formatter was added to this repo, and you're seeing it now because you've rebased.

https://github.com/openshift-helm-charts/development/actions/runs/6511492421/job/17687287216?pr=264#step:5:139

I've submitted a PR to fix it here: 

https://github.com/openshift-helm-charts/development/pull/274

komish (approved):
/lgtm

PR #265, opened 2023-10-06
Break down GitHub workflow into smaller jobs

OchiengEd:
Update GitHub workflow steps into smaller jobs


.github/workflows/build.yml
```diff
@@ -5,9 +5,14 @@ on:
...
 jobs:
-  chart-certification:
-    name: Chart Certification
+  setup:
+    name: Setup CI
     runs-on: ubuntu-22.04
+    outputs:
+      run_build: ${{ steps.check_build_required.outputs.run-build }}
+      verifier-action-image: ${{ steps.set-env.outputs.verifier-action-image }}
+      insecure_skip_tls_verify: ${{ steps.set-env.outputs.insecure_skip_tls_verify }}
```
komish, line 14: Can we normalize these outputs to use either hyphens (e.g. `run-build`) or camel case to divide words?

I'd say you don't have to modify the steps themselves now, because that will most likely require updating scripts. Leave those be, but when linking steps to outputs, lets normalize.

komish: Might go without saying, but remember that changing these outputs implies changing references in other steps.

.github/workflows/build.yml
```diff
@@ -89,24 +96,56 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    needs: [setup]
+
+    outputs:
+      report_content: ${{ steps.check_report.outputs.report_content }}
+      redhat_to_community: ${{ steps.check_report.outputs.redhat_to_community }}
+      message_file: ${{ steps.pr_comment.outputs.message-file }}
+      webCatalogOnly: ${{ steps.check_pr_content.outputs.webCatalogOnly }}
+      chartEntryName: ${{ steps.check_pr_content.outputs.chart-entry-name }}
+      chartNameWithVersion: ${{ steps.check_pr_content.outputs.chart-name-with-version }}
```
komish, line 111: Normalize output keys here too. Yes, I know I used camel case in my testing, but I don't care if we go hyphens or camel case.

komish (approved):
Just the two listed comments. Otherwise

/lgtm

.github/workflows/build.yml
OchiengEd: We can do this in a separate PR so since it entails updating the python scripts. I will put that on my plate

.github/workflows/build.yml
OchiengEd: I will do this in a separate pull request since it requires changing the Python scripts that are updating the variables

PR #267, opened 2023-10-10
Update actions/github-script version

OchiengEd:
Update actions/github-script package to v6 to migrate from `node12` to `node16`


komish (approved):
/lgtm

PR #271, opened 2023-10-12
Simplify output parameters

mgoerens:
The "chart-name-with-version" output parameter was only used to create the "tag" output parameter. This is now replaced by a single "release_tag" output parameter.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,7 +108,7 @@ jobs:
       message_file: ${{ steps.pr_comment.outputs.message-file }}
       webCatalogOnly: ${{ steps.check_pr_content.outputs.webCatalogOnly }}
       chartEntryName: ${{ steps.check_pr_content.outputs.chart-entry-name }}
-      chartNameWithVersion: ${{ steps.check_pr_content.outputs.chart-name-with-version }}
+      release_tag: ${{ steps.check_pr_content.outputs.release_tag }}
 
     steps:
       - name: Checkout
@@ -425,7 +425,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           REPORT_CONTENT: ${{ needs.chart-verifier.outputs.report_content }}
           CHART_ENTRY_NAME: ${{ needs.chart-verifier.outputs.chartEntryName }}
-          CHART_NAME_WITH_VERSION: ${{ needs.chart-verifier.outputs.chartNameWithVersion }}
           REDHAT_TO_COMMUNITY: ${{ needs.chart-verifier.outputs.redhat_to_community }}
           WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.webCatalogOnly }}
         id: release-charts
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

.github/workflows/build.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       message_file: ${{ steps.pr_comment.outputs.message-file }}
       webCatalogOnly: ${{ steps.check_pr_content.outputs.webCatalogOnly }}
       chartEntryName: ${{ steps.check_pr_content.outputs.chart-entry-name }}
-      chartNameWithVersion: ${{ steps.check_pr_content.outputs.chart-name-with-version }}
```
komish, line 111: Contextually, I do feel like "chart name with version" is a bit more clear. It's descriptive of the format of the data without showing the data.

With that said, I'd consider something like `chart_release_tag` to be acceptable. It's still descriptive, if only slight less so in terms of the content, but moreso in terms of its use.

Otherwise, I think I prefer it the way it is.

@jsm84 is out at the moment but if this is still open when they return, they can chime in.
@OchiengEd please feel free to add your thoughts as well.

.github/workflows/build.yml
```diff
@@ -439,11 +438,11 @@ jobs:
           cd ${CWD}
 
       - name: Release
-        if: ${{ steps.release-charts.outputs.tag != '' }}
```
komish, line 442: One thing to note here is that this step will transparently just _skip_ in cases where webCatalogOnly is an unexpected value (for whatever reason). I'm not sure how I feel about it.

With that said, I do very much prefer what you're changing this to as a criteria because it is way more descriptive.

This comment may be a no-op, as I'm not sure where I land on this. Feel free to disregard if it doesn't do anything to your opinion on what's being changed here (which is totally fine).

.github/workflows/build.yml
OchiengEd: I like how `chartNameWithVersion` is descriptive; you can picture the output without seeing its contents. Not opposed to renaming, but would recommend making it descriptive.

Lastly, all other variables are now using, snake case so the rename might help with bringing in consistency.

.github/workflows/build.yml
mgoerens: Let's agree the webCatalogOnly should never take another value than "True" or "False". Otherwise this is a bug that needs to be a) found, b) fixed in its time.

.github/workflows/build.yml
mgoerens: I would rather describe the usage rather than the content. `release_tag` describes the GitHub Tag that is used to cut the release of this chart. This variable / output parameter is only used for this purpose. (I wouldn't argue if we'd use the `chartNameWithVersion` in some other place for some other purpose that the GitHub release tag).

Also, `chartNameWithVersion` is actually not an completely accurate description of the format of the data. Its actually formatted as `tag_name = f"{organization}-{chart}-{version}"` so for instance `hashicorp-vault-0.25.0`. So maybe it should be `organizationWithChartNameAndVersion` ? :)

Naming is hard ...

OchiengEd: I suppose since it is only being used in a single location, we can afford to rename to `release_tag`.

mgoerens: I'll resolve the conversation. Thanks and sorry if that feels like bikeshedding !

(@komish It also makes sense how you see it and explain it)

mgoerens:
Fixed merge conflicts

OchiengEd (commented):
/lgtm

komish (approved):
/lgtm

PR #272, opened 2023-10-12
Add Python Code Style Enforcement to GitHub Actions

komish:
This PR adds a:

1) A Makefile within the scripts/ directory for use in enforcing style and managing virtualenvs
2) a GitHub Actions that consumes the Makefile targets to enforce style guidelines for changes to tooling within the scripts/ directory
3) An expanded .gitignore to remove a lot of cruft that gets added to the filesystem when building/installing various CI tools.

## Style

We've previously [applied Black](https://github.com/openshift-helm-charts/development/pull/247) to the code in this repository, and fixed a few pep8 complaints. This just takes that a bit further and starts to enforce it on PRs.

Linting is currently configured to auto-pass because we have a lot of things that are not automatically fixable (namely import positioning and line length). The idea here is that, when so motivated, we can review the CI results against the repository and go about fixing them as we go along (vs. requiring an immediate fix to all the line length complaints today).

`ruff` is used as the linter here simply due to its speed promises. If we have issues with it, we can eventually migrate to something else.

## Makefile

I make heavy use of venvs for python-related things. I'm less familiar with newer tooling thats in the same problem space. To that end, I've added Make targets that simply help facilitate the management and installation of virtualenvs. One target `venv.codestyle` is used for style tooling (`black`, `ruff`), the other `venv.tools` is used for our scripts themselves.

For development, there are additional "subtargets" that let you always reinstall these venvs, for cases where you're making changes to scripts and want to call them using the installed executables. E.g. `venv.tools.always-reinstall`.

## gitignore

Just adds various paths. This is borrowed from a partial of the GitHub Python default gitignore that's scaffolded for new projects.

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -1,13 +1,18 @@
 PY_BIN ?= python3
 
-VENV_CODESTYLE ?= venv.codestyle
+# The virtualenv containing code style tools.
+VENV_CODESTYLE = venv.codestyle
 VENV_CODESTYLE_BIN = $(VENV_CODESTYLE)/bin
 
-VENV_TOOLS ?= venv.tools
+# The virtualenv containing our CI scripts
+VENV_TOOLS = venv.tools
 VENV_TOOLS_BIN = $(VENV_TOOLS)/bin
 
 # This is what we pass to git ls-files.
 LS_FILES_INPUT_STR ?= 'src/*.py'
+
+.PHONY: default
+default: format lint
 
 # The same as format, but will throw a non-zero exit code
 # if the formatter had to make changes.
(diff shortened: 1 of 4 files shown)


scripts/Makefile
```diff
@@ -0,0 +1,56 @@
...
+		check \
+		$$(git ls-files $(LS_FILES_INPUT_STR))
+
+venv.tools:
+	$(MAKE) venv.tools.always-reinstall
+
+# This target will always install the tools at the venv.
+# Useful for development cases.
+.PHONY: venv.tools.always-reinstall
+venv.tools.always-reinstall:
```
OchiengEd, line 53: Any particular reason we would want to maintain different python virtual environments?

OchiengEd (approved):
/lgtm

scripts/Makefile
komish: The only reason is to keep the codestyle tooling separate, as it is not needed for deployment of all of our CI tooling that lives in the tools venv. That is, the tools venv is effectively what we install in the pipeline, and contains all of our entrypoints. That's effectively a "deployment", and the style tooling doesn't need to be there.

PR #275, opened 2023-10-16
Update GH actions from redhat-actions to use node 16

OchiengEd:
Update GitHub workflow actions from `redhat-actions` organization to versions that run on node 16.

This pull request:

* Updates `redhat-actions/openshift-tools-installer@v1` to `v1.12`
* Updates `redhat-actions/oc-login@v1` to `v1.2`
* Updates `redhat-actions/chart-verifier@v1.1` to `v1.2`

https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -226,7 +226,7 @@ jobs:
           API_SERVER=$( echo -n ${{ secrets.API_SERVER }} | base64 -d)
           echo "API_SERVER=${API_SERVER}" >> $GITHUB_OUTPUT
 
-      - uses: redhat-actions/oc-login@v1
+      - uses: redhat-actions/oc-login@v1.2
         id: oc_login
         if: ${{ steps.verify_requires.outputs.cluster_needed == 'true' }}
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -226,7 +226,7 @@ jobs:
           API_SERVER=$( echo -n ${{ secrets.API_SERVER }} | base64 -d)
           echo "API_SERVER=${API_SERVER}" >> $GITHUB_OUTPUT
 
-      - uses: redhat-actions/oc-login@v1.2
+      - uses: redhat-actions/oc-login@v1
         id: oc_login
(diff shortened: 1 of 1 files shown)


mgoerens (commented):
Thanks for this PR :+1:  That was bugging me too.

I'm wondering if we should pin minor releases. We could also just specify a major release.

That is:
* Keep `redhat-actions/openshift-tools-installer@v1` and `redhat-actions/oc-login@v1`: these were already using Node16
* Set `redhat-actions/chart-verifier@v1`

OchiengEd:
The big question would be do we value `predictability` or `convenience`? If using short version number e.g.  `redhat-actions/chart-verifier@v1` means we would be upgraded automatically through the minor versions of the GitHub action, then that subsequently implies we might see unexpected changes in behavior in the CI due to underlying changes in the GH action versions.

mgoerens:
I'd be in favor of sticking with major version pinning (i.e. `convenience`):
* Looking at the release notes for [oc-login](https://github.com/redhat-actions/oc-login/releases), [chart-verifier](https://github.com/redhat-actions/chart-verifier/releases), and [openshift-tools-installer](https://github.com/redhat-actions/openshift-tools-installer/releases), minor version bumps seem backward compatible. That's also in-line with SemVer guidance (`increment the MINOR version when you add functionality in a backward compatible manner`).
* In my limited experience with GitHub actions, I tend to see major version pinning (e.g. `actions/checkout@v4`)

If we decide to go for `predictability`, I would make sure that at least one of the core developers on this project subscribes to notifications for new releases on each of those GitHub actions. Otherwise it's too easy to forget :)

komish (approved):
/lgtm

komish:
@mgoerens @OchiengEd if I can comment on "predictability" vs. "convenience", I certainly land in the "predictability" camp for cases where we heavily rely on the component's core behavior. This is one of those cases. I'd also say it's less "predictability" and more "consistency".

**With that being said**, I (and we) kind of know for a fact here that the changes being made to this particular component are infrequent, and for this case, "convenience" wins.

Therefore: /lgtm

PR #269, opened 2023-10-11
Add acornett21 to OWNERS file to update workflows

komish:
Adds Adam to project OWNERS file so that they can contribute workflow changes.

Unblocks https://github.com/openshift-helm-charts/development/pull/268

PR #276, opened 2023-10-16
Extract repo path regex matching to a single library and reconfigure scripts to use it

komish:
In PR https://github.com/openshift-helm-charts/development/pull/269, we fixed an issue in the handling of plus sign values in our chart submission path (e.g. charts/partners/vendor/chartname/0.0.1+1/...) which was preventing us from successfully detecting valid files to modify/submit.

We fixed this in one location, but there are various locations that use nearly the same regex, some which did not include the exact same version support.

This PR extracts out the generation of that string to a library that can be called at any point.

The configurable keyword args were added simply to support the existing use cases in-repo (vs. altering those use cases). The goal here is to allow the library invocation to be compatible, and maybe iterate on how we handle matching these in the future (if needed).

### DIFF
--- a/scripts/src/release/releasechecker.py
+++ b/scripts/src/release/releasechecker.py
@@ -31,6 +31,7 @@ import semver
 import sys
 from release import release_info
 from release import releaser
+from reporegex import matchers
 
 sys.path.append("../")
 from owners import checkuser
@@ -38,7 +39,6 @@ from tools import gitutils
 from pullrequest import prartifact
 
 VERSION_FILE = "release/release_info.json"
-TYPE_MATCH_EXPRESSION = "(partners|redhat|community)"
 CHARTS_PR_BASE_REPO = gitutils.CHARTS_REPO
 CHARTS_PR_HEAD_REPO = gitutils.CHARTS_REPO
 DEV_PR_BASE_REPO = gitutils.DEVELOPMENT_REPO
@@ -69,7 +69,7 @@ def check_file_in_pr(api_url, pattern, error_value):
 def check_if_only_charts_are_included(api_url):
     print("[INFO] check if only chart files are included")
     chart_pattern = re.compile(
-        r"charts/" + TYPE_MATCH_EXPRESSION + "/([\w-]+)/([\w-]+)/.*"
+        matchers.submission_path_matcher(include_version_matcher=False) + r"./*"
     )
     return check_file_in_pr(api_url, chart_pattern, ERROR_IF_MATCH_NOT_FOUND)
 
@@ -77,7 +77,7 @@ def check_if_only_charts_are_included(api_url):
 def check_if_no_charts_are_included(api_url):
     print("[INFO] check if no chart files are included")
     chart_pattern = re.compile(
-        r"charts/" + TYPE_MATCH_EXPRESSION + "/([\w-]+)/([\w-]+)/.*"
+        matchers.submission_path_matcher(include_version_matcher=False) + r"./*"
     )
     return check_file_in_pr(api_url, chart_pattern, ERROR_IF_MATCH_FOUND)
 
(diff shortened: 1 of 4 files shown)


scripts/src/signedchart/signedchart.py
```diff
@@ -41,10 +42,12 @@ def get_verifier_flags(tar_file, owners_file, temp_dir):
...
     if api_url:
         files = prartifact.get_modified_files(api_url)
-        tgz_pattern = re.compile(r"charts/(\w+)/([\w-]+)/([\w-]+)/([\w\.-]+)/.*.tgz")
+        tgz_pattern = re.compile(
+            matchers.submission_path_matcher(strict_categories=False) + r".*.tgz"
+        )
         tgz_found = False
         prov_pattern = re.compile(
-            r"charts/(\w+)/([\w-]+)/([\w-]+)/([\w\.-]+)/.*.tgz.prov"
+            matchers.submission_path_matcher(strict_categories=False) + r".*.tgz.prov"
```
mgoerens, line 50: It's great that you paid attention to backward compatibility with this change. Part of me wonders if things break if we turn `strict_categories` to True here. I would imagine it doesn't and we can simplify this a little bit by getting rid of this argument.

mgoerens (changes requested):
Minor change requeted and a couple of questions / comments

scripts/src/signedchart/signedchart.py
komish: I had the same thought but didn’t want to find out 😂 in my mind, it would seem like having less strict category could be a thing in some  circumstances, like if a contributor were to typo a category some how… but I didn’t chase it down. I can certainly try it.

komish: I might aim to merge this as is so I can include  this in the release because of the fact that the current changes do not solve the plus sign versioning issue sufficiently, which this is trying to help.

scripts/src/reporegex/matchers.py
```diff
@@ -0,0 +1,36 @@
...
+    Callers should append any relevant path matching to the end of the string
+    returned from this function. E.g. "/.*"
+
+    Keyword Args
+        base_dir -- The base path of the expression statement. Should
+            not be empty.
+        strict_categories -- Whether the category matcher should match only the
+            relevant categories, or any word at all.
+        include_version_matcher -- Whether or not the version matcher should be appended.
+            In some cases, the caller of this regex doesn't care about the versioning detail.
```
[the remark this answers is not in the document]
komish: I’ll take a look here shortly. I’m strongly dissatisfied with the many forms of docstrings 😞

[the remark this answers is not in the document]
komish: This docstring has been updated to match the guidelines as of commit https://github.com/openshift-helm-charts/development/commit/99687e35e74e162913c36f6d14e910f330d05f7c

Feel free to mark this resolved if/when you're satisfied with the way this looks.

scripts/src/release/releasechecker.py
```diff
@@ -69,15 +70,15 @@ def check_file_in_pr(api_url, pattern, error_value):
 def check_if_only_charts_are_included(api_url):
     print("[INFO] check if only chart files are included")
     chart_pattern = re.compile(
-        r"charts/" + TYPE_MATCH_EXPRESSION + "/([\w-]+)/([\w-]+)/.*"
```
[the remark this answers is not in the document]
komish: I stripped out TYPE_MATCH_EXPRESSIONS as I saw no other references to it. Done in commit https://github.com/openshift-helm-charts/development/commit/99687e35e74e162913c36f6d14e910f330d05f7c

Please mark resolved if/when you're satisfied with this :+1: . Good catch!

scripts/src/signedchart/signedchart.py
komish: I think I landed on wanting this type of evaluation to happen in a later change. If that's workable for you, please mark this resolved.

mgoerens: Absolutely fine with me !

mgoerens (approved):
LGTM

PR #277, opened 2023-10-17
Configure more locations to use the submission regex matchers

komish:
The most recent PR to update our use of submission path regex missed a couple of locations where this was being used, so this PR should address those locations.

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -23,6 +23,7 @@ from report import report_info
 from report import verifier_report
 from signedchart import signedchart
 from pullrequest import prartifact
+from reporegex import matchers
 from tools import gitutils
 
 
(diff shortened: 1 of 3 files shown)


scripts/src/chartprreview/chartprreview.py
```diff
@@ -58,7 +59,7 @@ def get_modified_charts(directory, api_url):
     """
     print("[INFO] Get modified charts. %s" % directory)
     files = prartifact.get_modified_files(api_url)
-    pattern = re.compile(r"charts/(\w+)/([\w-]+)/([\w-]+)/([\w\.\-+]+)/.*")
+    pattern = re.compile(matchers.submission_path_matcher() + r"/.*")
```
mgoerens, line 62: To be consistent we should use `strict_categories=False` here.

mgoerens (changes requested):

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -24,12 +24,24 @@
...
 def get_modified_charts(api_url):
+    """Reads a PR's modified files to get submission details and exits if none are found.
+
+    Args:
+        api_url: URL to the API representation of a pull request.
+          E.g. https://api.github.com/repos/openshift-helm-charts/charts/pulls/1
+
+    Returns:
+        The category, organization, chart name, and version.
+    """
```
mgoerens, line 40: Would you be fine with removing this docstring ? See https://github.com/openshift-helm-charts/development/pull/278

komish: Fixed

OchiengEd:
No further recommendations. Looks good

OchiengEd (approved):
/lgtm

PR #279, opened 2023-10-18
Release 1.6.0

komish:
Quite a few changes landing here, including workflows, scripts, etc. Bumping the Y-stream number for these reasons.


komish:
The only step that seems to have failed is the merge step. E2E testing passed, so for that reason, I'll merge this in. 
PRs to stage/charts are experiencing unrelated failures, to be investigated.

PR #283, opened 2023-10-23
Improve markdown comment formatting library and update PR comment format

komish:
This is the first PR in an effort to restructure how we handle PR comments so it's a bit more reliable in our pipeline. This PR mostly just cleans up the library logic, and produces a formatted PR comment.

We currently don't have a templating dependency like Jinja, so I continued to leave any templating dependency out. For now. That might change in the future.

With that said, this PR adds helper functionality **append_to** that helps build the comment, and consolidates our use of multi-line strings (which visually breaks the flow of the source code), and manually managed line breaks to as few places as possible.

This PR also adds a 'Task Table' to the PR comment, which reflects what GitHub Action steps were considered in the production of the comment. We could make this more useful over time. Admittedly, this is mostly useful for us as maintainers, to review what broke at a glance.

Finally, I've added comments to clarify intent in the functionality, and have added print statements to allow us to review the generated comment in CI should it not get posted to the PR itself (for any reason, like GHA outages).

### DIFF
--- a/scripts/src/pullrequest/prepare_pr_comment.py
+++ b/scripts/src/pullrequest/prepare_pr_comment.py
@@ -28,8 +28,13 @@ def get_verifier_errors_comment():
 
 
 def get_verifier_errors_trailer():
-    return "Please create a valid report with the [chart-verifier](https://github.com/redhat-certification/chart-verifier) \
-and ensure all mandatory checks pass."
+    return " ".join(
+        [
+            "Please create a valid report with the",
+            "[chart-verifier](https://github.com/redhat-certification/chart-verifier)",
+            "and ensure all mandatory checks pass.",
+        ]
+    )
 
 
 def get_look_at_job_output_comment():
@@ -38,60 +43,63 @@ link next to "CI / Chart Certification" job status towards the end of this page.
 
 
 def prepare_failure_comment():
-    msg = f"""\
-{get_failure_comment()}
-{get_look_at_job_output_comment()}"""
+    """assembles the comment for certification failures
+
+    Will attempt to read a file with error messaging from the filesystem
+    and includes that information in its content. (e.g. ./pr/errors)
+    """
+    msg = get_failure_comment()
+    msg = append_to(msg, get_look_at_job_output_comment())
     if os.path.exists("./pr/errors"):
         errors = open("./pr/errors").read()
-        msg += f"""
-{get_verifier_errors_comment()}
-
-{errors}
-
-{get_verifier_errors_trailer()}
-
-"""
+        msg = append_to(msg, get_verifier_errors_comment())
+        msg = append_to(msg, errors)
+        msg = append_to(msg, get_verifier_errors_trailer())
         gitutils.add_output("error-message", errors)
     else:
         gitutils.add_output("error-message", get_failure_comment())
     return msg
 
 
-def prepare_success_comment():
-    msg = f"{get_success_coment()}.\n\n"
-    return msg
-
-
 def prepare_pr_content_failure_comment():
-    msg = f"{get_content_failure_message()} \n"
+    """Generate a message for PR Content Check Failures
+
+    This function reaches into the environment for known variables
+    that contain error messages.
+
+    Error messages are then passed into the GITHUB_OUTPUT.
+
+    Returns a formatted string containing error message contents.
+    """
+    msg = f"{get_content_failure_message()}"
     pr_content_error_msg = os.environ.get("PR_CONTENT_ERROR_MESSAGE", "")
     owners_error_msg = os.environ.get("OWNERS_ERROR_MESSAGE", "")
     if pr_content_error_msg:
         gitutils.add_output("error-message", pr_content_error_msg)
-        msg += f"{pr_content_error_msg}\n\n"
+        msg = append_to(msg, f"{pr_content_error_msg}")
     if owners_error_msg:
         gitutils.add_output("error-message", owners_error_msg)
-        msg += f"{owners_error_msg}\n\n"
+        msg = append_to(msg, f"{owners_error_msg}")
     return msg
 
 
 def prepare_run_verifier_failure_comment():
     verifier_error_msg = os.environ.get("VERIFIER_ERROR_MESSAGE", "")
     gitutils.add_output("error-message", verifier_error_msg)
-    msg = f"""
-{verifier_error_msg}
-
-{get_look_at_job_output_comment()}
-"""
+    msg = verifier_error_msg
+    msg = append_to(msg, get_look_at_job_output_comment())
     return msg
 
 
 def prepare_community_comment():
-    msg = f"{get_community_review_message()}\n\n"
+    msg = f"{get_community_review_message()}"
     if os.path.exists("./pr/errors"):
         errors = open("./pr/errors").read()
-        msg += "However, please note that one or more errors were found while building and verifying your pull request:\n\n"
-        msg += f"{errors}\n\n"
+        msg = append_to(
+            msg,
+            "However, **please note** that one or more errors were found while building and verifying your pull request:",
+        )
+        msg = append_to(msg, f"{errors}")
     return msg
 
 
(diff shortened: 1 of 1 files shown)


komish:
Note to reviewers: A lot of the work here was cleaning up how the library worked, but note that the PR comments now look different, and that's somewhat transparent. Click through the PRs CI made to see what I mean.

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -38,138 +43,245 @@ def get_look_at_job_output_comment():
...
+        detail_message = append_to(detail_message, get_success_coment())
         gitutils.add_output("pr_passed", "true")
     else:  # Handle various failure scenarios.
         if pr_content_result == "failure":
-            msg += prepare_pr_content_failure_comment()
+            detail_message = append_to(
+                detail_message, prepare_pr_content_failure_comment()
+            )
             gitutils.add_output("pr_passed", "false")
         elif run_verifier_result == "failure":
```
mgoerens, line 240: We should consider displaying error messages for each failure that was received rather than only for the first one.

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -38,138 +43,245 @@ def get_look_at_job_output_comment():
...
         elif oc_install_result == "failure":
-            msg += prepare_oc_install_fail_comment()
+            detail_message = append_to(
+                detail_message, prepare_oc_install_fail_comment()
+            )
             gitutils.add_output("pr_passed", "false")
         else:
-            msg += prepare_generic_fail_comment()
+            detail_message = append_to(detail_message, prepare_generic_fail_comment())
             gitutils.add_output("pr_passed", "false")
```
mgoerens, line 260: This is maybe out of scope, but looking at [this sandbox PR](https://github.com/openshift-helm-charts/sandbox/pull/71823), the message is not correct.

chart- verifier was skipped, outcome is Failed, we get a generic error message, but the PR is successfully merged.

mgoerens (changes requested):

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -38,138 +43,245 @@ def get_look_at_job_output_comment():
...
+    It's intended to allow us to use a multi-line string but work entirely at
+    the same indentation level in source code.
+
+    Args:
+        msg: The original message to which we should append new_content.
+        new_content: the new string to add. use_horizontal_divider: Whether to
+        divide the content
+          with a horizontal line (in markdown.) Horizontal lines are surrounded
+          in newlines to ensure that it doesn not inadvertently cause preceding
+          content to become a Header.
```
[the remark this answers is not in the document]
komish: My auto-wrapping extension breaks docstrings for some reason. Good catch.

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -38,138 +43,245 @@ def get_look_at_job_output_comment():
...
+
+
+def get_support_information():
+    reqs_doc_link = "https://github.com/redhat-certification/chart-verifier/blob/main/docs/helm-chart-checks.md#types-of-helm-chart-checks"
+    support_link = "https://redhat-connect.gitbook.io/red-hat-partner-connect-general-guide/managing-your-account/getting-help/technology-partner-success-desk"
+    return "\n".join(
+        [
+            "For information on the certification process see:",
+            f"- [Red Hat certification requirements and process for Kubernetes applications that are deployed using Helm charts.]({reqs_doc_link}).",
+            f"- For support, connect with our [Technology Partner Acceleration Desk]({support_link}).",
```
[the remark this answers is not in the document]
komish: I had switched the link over to the Access article but must have undid that change. Good catch.

scripts/src/pullrequest/prepare_pr_comment.py
komish: Definitely worth considering, but I think I'd like to defer this to a future PR.

scripts/src/pullrequest/prepare_pr_comment.py
komish: Looking at this before I push fixes out.

komish: Our logic is looking for an explicit failure, but in some cases we don't run the `install-oc` task and that shouldn't be considered a failure.

For now, I'm consolidating changes to this file but I think we may have opportunities for this to be fixed in a better way. As I shift to looking at the workflows (vs. this file), I may change this up. For now, I've allowed `"skipped"` and `False` as valid values for the OC installation, and `"skipped"` as a valid value for the run-verifier task because it may not be needed in all cases.

scripts/src/pullrequest/prepare_pr_comment.py
```diff
@@ -38,138 +43,245 @@ def get_look_at_job_output_comment():
...
+    template. msg and new_content are joined with either a newline separator, or
+    a horizontal line (if the keyword argument is provided).
+
+    It should be used in cases where the caller needs to join semi-related
+    ideas. Callers should instead use the join string method in cases where the
+    msg being constructed is a part of the same 'idea', or 'paragraph'.
+
+    Developer Note: The use of the strip method is a developer optimization.
+    It's intended to allow us to use a multi-line string but work entirely at
+    the same indentation level in source code.
```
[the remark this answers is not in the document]
komish: I removed the developer note and instead added a comment explicitly where `strip()` is called.

The gist here is that multi-line strings preserve whitespace. I'm adding the multiline string instantiator and terminator (e.g. `"""`) on new lines, which would artificially add newlines in the string itself. This is just to keep the source code easy to read, and so I strip them out using `strip()` so they don't end up in our result.

If it's too unclear, and you prefer me to just skip using strip and change the source, I can do that.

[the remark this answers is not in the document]
mgoerens: Thanks for the explanation. It's clear now.

I was mostly thrown off by the comment regarding indentation level in the source code. We do need to kind of mess up the indentation when working with multi-line string (as the content of this multi-line string cannot be indented with the rest of the code.) I interpreted your initial comment as trying to solve this issue, which confused me.

I understand now that the intention is to keep the same indentation level "inside" the multi-line string content i.e. avoid doing:

```
    return f"""{msg}
{divider_string}
{new_content}"""
```

I'm okay with using `strip()` here, and we should leave your comment.

Just note that this will also strip out any leading whitespace in the `msg` string. I don't think that's an issue.

scripts/src/pullrequest/prepare_pr_comment.py
mgoerens: Could you explain, and maybe add a comment in the code, why "False" is a valid result for OC install ?

komish: I should clarify: I don't like the use of False here. However, I believe its use is to be a third option other than passed or failed that isn't explicitly passed or failed. In other words, it doesn't trigger pass logic, and doesn't trigger fail logic, it just is. I would assume False is used to help with any logic cases where I need to say `if oc_install_result`. None may also be valid in that case. Either way, it was originally false, so I left it as is for the moment.

mgoerens: Ok... A few things don't make complete sense to me, but those issues where already present prior to your PR.

I'll open a separate issue to discuss those.

scripts/src/pullrequest/prepare_pr_comment.py
mgoerens: Opened https://github.com/openshift-helm-charts/development/issues/286 to track that improvement

scripts/src/pullrequest/prepare_pr_comment.py
mgoerens: Opened #287 to handle separate issues I found regarding `oc-install`

PR #288, opened 2023-10-26
Better handle oc-install outcome

### LINKED ISSUE #287: Improve handle of oc-install outcome
Comes from [this discussion](https://github.com/openshift-helm-charts/development/pull/283#discussion_r1369898088)

Currently, the step that prepares the PR comment uses the `oc-install.conclusion` value to determine if this step was successful, skipped, or failed.

This step doesn't `continue-on-error`, therefore it makes little sense to check its `conclusion`. Here we should instead use `outcome`. See https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context.

Furthermore, the way this argument is currently provided to the python script makes it possible for this value to be a boolean set to `False` instead of a string.

mgoerens:
The oc-install step doesn't continue-on-error, hence we should use .outcome instead of .conclusion to check if it runned successfully, was skipped or failed.

This information is never empty, hence there is no need to default the value of this argument to False in prepare_pr_comment.py.

Closes #287


OchiengEd (approved):
/lgtm

komish (approved):
/lgtm

PR #289, opened 2023-10-27
Consolidate Github release management in one step

mgoerens:
~Please do not review/merge just yet, I want to see the pipeline results before testing/applying further code optimizations~ READY !

So far the chart-repo-manager script had multiple functions:
* Create the GitHub Release (only if the chart's sources are provided)
  using chart-releaser.
* Prepare the index entry for this chart.
* Update the Helm repository index.

In addition, a separate step in our build.yaml workflow added the
report.yaml to the GitHub release (potentially creating it, in the case
the chart's sources are not provided).

This commit brings the following changes:
* Reduce the role of the chart-repo-manager script to a) the preparation
  of all artifacts that go into a GitHub Release, and b) the crafting of
  the index entry.
* Consolidate the GitHub release creation and the upload of all its
  artifacts into a single step. This eliminates the need to use
  chart-releaser.
* Create a dedicated step for the update of the Helm repository index.
* Favor running shell commands directly in the steps 'run:' rather than
  in the python scripts, using subprocesses. This is expecially true for
  the multiple "git" commands used to update the index.

This makes https://github.com/openshift-helm-charts/charts/issues/1080
obsolete and can be closed.

---

Note for reviewers: To be discussed but there is room for further improvement, for instance:
- The step that updates the index is quite large due mostly to the use of shell commands over python subprocesses, and could be split into smaller more readable steps.
- We could reconsider the need to create a worktree vs checkout the index branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -461,43 +461,77 @@ jobs:
           git config --global user.name "github-actions[bot]"
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
-      - name: Release Charts
+      - name: Prepare Chart Release and index entry
         if: ${{ needs.setup.outputs.run_build == 'true' }}
         env:
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
-          GITHUB_REF: ${{ github.ref }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           REPORT_CONTENT: ${{ needs.chart-verifier.outputs.report_content }}
-          CHART_ENTRY_NAME: ${{ needs.chart-verifier.outputs.chart_entry_name }}
           REDHAT_TO_COMMUNITY: ${{ needs.chart-verifier.outputs.redhat_to_community }}
           WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.web_catalog_only }}
           OCP_VERSION_RANGE: ${{ steps.get-ocp-range.outputs.ocp-version-range }}
-        id: release-charts
+        id: prepare-chart-release
         run: |
-          tar zxvf ./scripts/dependencies/helm-chart-releaser/chart-releaser_1.2.0_linux_amd64.tar.gz
-          sudo cp -f cr /usr/local/bin/cr
-          INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "refs/heads/gh-pages"; else echo "${GITHUB_REF}-gh-pages"; fi)
-          CWD=`pwd`
           cd pr-branch
           ../ve1/bin/chart-repo-manager \
             --repository=${{ github.repository }} \
-            --index-branch=${INDEX_BRANCH} \
             --api-url=${{ github.event.pull_request._links.self.href }} \
-            --pr-number=${{ github.event.number }}
-          cd ${CWD}
 
-      - name: Release
+      # Upload the report file, potentially paired with a public key and, if provided, the chart's tarball and its prov file.
+      # Only the report file is always included.
+      # The release tag format is <organization_name>-<chart_name>-<chart_version>
+      - name: Create GitHub release
         if: ${{ needs.chart-verifier.outputs.web_catalog_only == 'False' }}
-        uses: softprops/action-gh-release@v0.1.12
-        continue-on-error: true
+        uses: softprops/action-gh-release@v0.1.15
         with:
           tag_name: ${{ needs.chart-verifier.outputs.release_tag }}
           files: |
-              ${{ steps.release-charts.outputs.report_file }}
-              ${{ steps.release-charts.outputs.public_key_file }}
+              ${{ steps.prepare-chart-release.outputs.report_file }}
+              ${{ steps.prepare-chart-release.outputs.public_key_file }}
+              ${{ steps.prepare-chart-release.outputs.path_to_chart_tarball }}
+              ${{ steps.prepare-chart-release.outputs.prov_file_name }}
           fail_on_unmatched_files: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update Helm repository index
+        if: ${{ needs.setup.outputs.run_build == 'true' }}
+        env:
+          CHART_ENTRY_NAME: ${{ needs.chart-verifier.outputs.chart_entry_name }}
+          WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.web_catalog_only }}
+        run: |
+          INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "gh-pages"; else echo "${GITHUB_REF##*/}-gh-pages"; fi)
+
+          echo "[INFO] Creating Git worktree for index branch"
+          INDEX_DIR=$(mktemp -d)
+          git remote add upstream "$GITHUB_SERVER_URL"/"$GITHUB_REPOSITORY"
+          git fetch upstream "$INDEX_BRANCH"
+          git worktree add --detach "$INDEX_DIR" upstream/"$INDEX_BRANCH"
+
+          if [ "$WEB_CATALOG_ONLY" == "True" ]; then
+            INDEX_FILE="unpublished-certified-charts.yaml"
+          else
+            INDEX_FILE="index.yaml"
+          fi
+
+          source ve1/bin/activate
+          cd $INDEX_DIR
+          update-index \
+            --index-branch=${INDEX_BRANCH} \
+            --index-file=${INDEX_FILE} \
+            --repository=${{ github.repository }} \
+            --chart-entry='${{ steps.prepare-chart-release.outputs.chart_entry }}' \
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -461,44 +461,77 @@ jobs:
...
-          cd ${CWD}
 
-      - name: Release
+      # Upload the report file, potentially paired with a public key and, if provided, the chart's tarball and its prov file.
+      # Only the report file is always included.
+      # The release tag format is <organization_name>-<chart_name>-<chart_version>
+      - name: Create GitHub release
         if: ${{ needs.chart-verifier.outputs.web_catalog_only == 'False' }}
-        uses: softprops/action-gh-release@v0.1.12
-        continue-on-error: true
```
komish, line 492: We no longer want to continue on error?

.github/workflows/build.yml
```diff
@@ -461,44 +461,77 @@ jobs:
...
+        uses: softprops/action-gh-release@v0.1.15
         with:
           tag_name: ${{ needs.chart-verifier.outputs.release_tag }}
           files: |
-              ${{ steps.release-charts.outputs.report_file }}
-              ${{ steps.release-charts.outputs.public_key_file }}
+              ${{ steps.prepare-chart-release.outputs.report_file }}
+              ${{ steps.prepare-chart-release.outputs.public_key_file }}
+              ${{ steps.prepare-chart-release.outputs.path_to_chart_tarball }}
+              ${{ steps.prepare-chart-release.outputs.prov_file_name }}
```
komish, line 491: I'm assuming, but I also want to confirm - If the outputs referenced here are missing, does it just resolve to a "blank line" and this action doesn't publish that artifact?

.github/workflows/build.yml
```diff
@@ -461,44 +461,77 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Update Helm repository index
+        if: ${{ needs.setup.outputs.run_build == 'true' }}
+        env:
+          CHART_ENTRY_NAME: ${{ needs.chart-verifier.outputs.chart_entry_name }}
+          WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.web_catalog_only }}
+        run: |
+          INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "gh-pages"; else echo "${GITHUB_REF##*/}-gh-pages"; fi)
```
komish, line 502: Just a note: this seems like one of those metadata items that we can derive and set workflow-wide.
No action required for this comment. Just wanted to mention it so it's on your radar for the issue on the same topic. (Feel free to "Resolve Conversation" once read).

.github/workflows/build.yml
```diff
@@ -461,44 +461,77 @@ jobs:
...
+
+          source ve1/bin/activate
+          cd $INDEX_DIR
+          update-index \
+            --index-branch=${INDEX_BRANCH} \
+            --index-file=${INDEX_FILE} \
+            --repository=${{ github.repository }} \
+            --chart-entry='${{ steps.prepare-chart-release.outputs.chart_entry }}' \
+            --chart-url="${{ steps.prepare-chart-release.outputs.chart_url }}" \
+            --version="${{ steps.prepare-chart-release.outputs.version }}"
```
komish, line 524: This is one of those CI-isms - should these values be inputs to the step (e.g. as `env` definitions) or called directly here?

Making these inputs to the step makes it a bit more portable, and I do think this step will benefit from portability, given that I'd like to be able to re-run index modification work in the future in cases where steps have failed transiently.

mgoerens: So far I considered a bad practice to extensively use `os.environ.get` within our python script to pass what essentially are simple arguments for this scripts. Sometimes those calls to `os.environ.get` are made deep inside some function calls, which can make it really cryptic what environment variables are actually used and where.

Hence, when creating this new `update-index` script (though it's mostly code shifted away from `chart-repo-manager`), I avoided using environment variables and preferred `argparse`.

But... it seems you're not actually commenting on this, rather you're suggesting to have an environment variable, that is then used as argument to the python script ? At the moment I fail to see how that makes the step more "portable", but I'm happy to move to this kind of design if you prefer.

.github/workflows/build.yml
mgoerens: Good idea, I'll keep this in mind for a future iteration on this.

.github/workflows/build.yml
mgoerens: Yes that is correct. The report file is the only file always included in the release.

.github/workflows/build.yml
mgoerens: Good point (BTW I'm not always happy with the way we extensively use `continue-on-error` and fail later. I think it makes debugging more complicated and is not always necessary. Maybe a design discussion we need to have, I'll try to document my annoyance in a separate issue one day :) ).

In this particular case I don't see the benefit of `continue-on-error`. Let's take a look at the following steps:
- `Update Helm repository index` will fail anyway if the release doesn't exist. There is a check within the `update-index` script that the release exists for this chart.
- `Retrieve PR comment for metrics` and `Add metrics` have the `always()` condition, so they will run no matter what.

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -401,204 +311,6 @@ def create_index_from_report(category, ocp_version_range, report_path):
...
-                "Found an integrity issue. SHA256 digest passed does not match SHA256 digest computed."
-            )
-    elif not pkg_digest:
-        # Digest was not passed and could not be computed
-        raise Exception(
-            "Was unable to compute SHA256 digest, please ensure chart url points to a chart package."
-        )
-
-
-def update_index_and_push(
```
komish, line 440: For clarity - this logic is what's being replaced by the new step in the workflow, correct?

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -401,204 +311,6 @@ def create_index_from_report(category, ocp_version_range, report_path):
     return chart_entry
 
 
-def set_package_digest(chart_entry):
```
komish, line 404: I might be missing it - is this being calculated and added in the replacement logic that handles index updates?

scripts/src/chartrepomanager/chartrepomanager.py
mgoerens: Yes, the new step can (and should probably) be divided into three smaller logical steps:
* Preparing the `git worktree` -> This was done at the beginning of `chartrepomanager` but only used here. It belongs to the "update index" logic. As these are plain git commands, I moved them to the shell.
* Downlading and updating the index data -> This was part of this function and is now the `update-index` script. This is actually the only part that remained as python code.
* Pushing to the index branch -> This was also part of this function. As they are plain git commands, I moved them to the shell.

.github/workflows/build.yml
komish: CI didn't fail, so we can continue without it for now. Typically, the reason I see it is so that we can comment on the PR, but at this point in the workflow, I don 't think there's any commenting that comes later so we should be in the  clear.

komish:
@mgoerens one last question here (aside from those already in-lined):

I see this log line being emitted in the step where the github release and index entries are being _prepared_ (read: not created).

> [INFO] Helm package was released at https://github.com/openshift-helm-charts/sandbox/releases/download/hashicorp-25bc8561e1b34b06985003c84efe4963-289-vault-0.17.0/vault-0.17.0.tgz

https://github.com/openshift-helm-charts/sandbox/actions/runs/6810139895/job/18517922140#step:9:195

Are _we_ emitting this, a tool we're calling, or is this a remnant from when we called chart-releaser?

komish (approved):
/lgtm

.github/workflows/build.yml
mgoerens: Understood. I think it makes sense.

I will Resolve for now and keep this in mind for future iterations. It's unfortunately not that trivial here because `${{ steps.prepare-chart-release.outputs.chart_entry }}` holds a dict that is loaded using `json.loads` in `argparse`, and it needs to be single-quoted in order to work. But if I use a environment variable it won't be evaluated because of the single quotes.

```python
    parser.add_argument(
        "-e",
        "--chart-entry",
        dest="chart_entry",
        type=json.loads,
        required=True,
        help="Index entry to add",
    )
```

It may not be a very good idea. It's actually the first time I use `json.loads` in as a argparse type. I could write to a file, or use base64 encoding/decoding instead... As we don't consider this blocking I'll keep this for a later iteration.

mgoerens:
@komish Good catch, I will removed it as in the next step "Create GitHub release" we also have the URL for this release printed out, e.g.
```
🎉 Release ready at https://github.com/openshift-helm-charts/sandbox/releases/tag/hashicorp-25bc8561e1b34b06985003c84efe4963-289-vault-0.17.0
```

I also doubled check the other INFO/DEBUG messages and I added one missing INFO message was previously present:
```
          echo "[INFO] Add and commit changes to git"
```
in build.yaml before pushing the changes to the index branch.

komish (approved):
/lgtm

PR #290, opened 2023-11-03
Make the PR comment accessible to the metrics step

komish:
At the moment, Metrics reporting is failing because the PR comment it needs is no longer available in the job where it runs.

https://github.com/openshift-helm-charts/charts/actions/runs/6709583883/job/18233057391#step:11:53

I believe that the metrics step can probably be extracted out into its own execution such that it doesn't cause the release step to fail, but until that gets prioritizes, this PR pulls the PR comment contents from the `chart-verifier` task and into the `release` task by encoding it in a step of the former job, and then writing it back to the expected location in a step of the latter job.

Definitely a temporary measure, but it's a quick way to get the release tasks to start passing. At the moment, I would expect them all to fail, even though all the actual releasing activity (e.g. pushing charts, public keys, indexing) succeeded.

Once this merges, I'll probably cut a release to get this into the production pipeline, even though the content is small.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,6 +106,7 @@ jobs:
       report_content: ${{ steps.check_report.outputs.report_content }}
       redhat_to_community: ${{ steps.check_report.outputs.redhat_to_community }}
       message_file: ${{ steps.pr_comment.outputs.message-file }}
+      message_text_base64: ${{ steps.encode_pr_comment.outputs.message-text-base64 }}
       web_catalog_only: ${{ steps.check_pr_content.outputs.web_catalog_only }}
       chart_entry_name: ${{ steps.check_pr_content.outputs.chart-entry-name }}
       release_tag: ${{ steps.check_pr_content.outputs.release_tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -340,6 +341,15 @@ jobs:
         run: |
           ve1/bin/pr-comment ${{ steps.check_pr_content.outcome }} ${{ steps.run-verifier.outcome }} ${{ steps.check_report.conclusion }}
 
+      # Note(komish): This step is a temporary fix for the metrics step in the next job
+      # which expects the PR comment to exist at the specified filesystem location.
+      - name: Encode PR Comment for Metrics
+        id: encode_pr_comment
+        if: ${{ always() && needs.setup.outputs.run_build == 'true' }}
+        run: |
+          commentBase64=$(base64 --wrap=0 ./pr/comment)
```
mgoerens, line 350: Use `${{ steps.pr_comment.outputs.message-file }}` instead of hard-coded `./pr/comment`

komish: Done!

mgoerens (approved):

PR #298, opened 2023-11-15
Use env vars rather than step's context

mgoerens:
This change is meant to:
a) Make it easier to spot which inputs needs to be passed to a step.
b) Improve step's portability.

This is a first iteration of this logic, focusing on the recent changes to chart-repo-manager and update-index.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -469,12 +469,13 @@ jobs:
           REDHAT_TO_COMMUNITY: ${{ needs.chart-verifier.outputs.redhat_to_community }}
           WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.web_catalog_only }}
           OCP_VERSION_RANGE: ${{ steps.get-ocp-range.outputs.ocp-version-range }}
+          API_URL: ${{ github.event.pull_request._links.self.href }}
         id: prepare-chart-release
         run: |
           cd pr-branch
           ../ve1/bin/chart-repo-manager \
-            --repository=${{ github.repository }} \
-            --api-url=${{ github.event.pull_request._links.self.href }} \
+            --repository="$GITHUB_REPOSITORY" \
+            --api-url="$API_URL" \
 
       # Upload the report file, potentially paired with a public key and, if provided, the chart's tarball and its prov file.
       # Only the report file is always included.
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -469,12 +469,13 @@ jobs:
...
           WEB_CATALOG_ONLY: ${{ needs.chart-verifier.outputs.web_catalog_only }}
           OCP_VERSION_RANGE: ${{ steps.get-ocp-range.outputs.ocp-version-range }}
+          API_URL: ${{ github.event.pull_request._links.self.href }}
         id: prepare-chart-release
         run: |
           cd pr-branch
           ../ve1/bin/chart-repo-manager \
-            --repository=${{ github.repository }} \
-            --api-url=${{ github.event.pull_request._links.self.href }} \
+            --repository="$GITHUB_REPOSITORY" \
```
komish, line 477: Initially, I think I intended for this to be `${{ env.GITHUB_REPOSITORY }}`, but I think what you've done here makes sense, given the original goal I had in mind.

Looking at this now, I wonder if there are any downsides to using shell-specific environment variables here. I can't think of any - is there anything you can think of?

.github/workflows/build.yml
```diff
@@ -516,12 +520,12 @@ jobs:
...
-            --repository=${{ github.repository }} \
-            --chart-entry="${{ steps.prepare-chart-release.outputs.chart_entry }}" \
-            --chart-url="${{ steps.prepare-chart-release.outputs.chart_url }}" \
-            --version="${{ steps.prepare-chart-release.outputs.version }}"
+            --index-branch="$INDEX_BRANCH" \
+            --index-file="$INDEX_FILE" \
+            --repository="$GITHUB_REPOSITORY" \
+            --chart-entry="$CHART_ENTRY" \
+            --chart-url="$CHART_URL" \
+            --version="$VERSION"
```
komish, line 528: As a general practice, I would prefer if these should all be enclosed in brackets. E.g. `$INDEX_BRANCH` should be `${INDEX_BRANCH}` as it was before, and the remaining environment variables should match.

Good explanation here: https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables.

It's [my] preference and convention, not a requirement, and so I'll leave it to your discretion to address this feedback given that this case isn't affected by the lack of curly braces.

komish:
@mgoerens I'm actually going to close this for now.

In the PoC I'm working on, I noticed that using environment variables in `run` steps leaves it unresolved within the code. E.g.

```
run: echo "${FOO}"
```

...would look exactly like that, at execution, in logs. It's not hard to resolve (environment variables are visible in all action run logs), but it's still an extra step. If we use the `${{ ... }}` syntax, the log would have the resolved value.

E.g.

```
run: command --flag=${{ steps.stepname.outputs.myvalue }}
```

... would resolve to

```
run: command --flag=thevalue
```

That reads a little bit easier.

The original goal was to be able to just "copy paste" the script definition in our `run` blocks into a bash, python, javascript, etc. script - and be able to run it. That's what I meant by "more portable". I think at this moment, though, I'd rather see it resolved instead of having to look at an extra place in the log.

Feel free to re-open this, but given that I asked for this initially, I'll mark it closed now. Thanks again for you work.

PR #299, opened 2023-11-15
Move to pyproject.toml and bump to Python 3.10

komish:
Couple of things happening here. Most of this is coming from a side-effect of noticing that Python 3.12 doesn't play nicely with our script installation.

- `setup-python` tasks at v2 were bumped to v4. We already used v4 in many places, but some places didn't have v4.
- **setup.py** is to be removed (after this merges, see note below) in favor of **pyproject.toml**. It seems setup.py-based installations are deprecated. I've elected not to introduce alternative tooling to manage packaging, but we can look into it in the future.
- The installation of our local package, which referenced setup.py, has been updated to use `pip`.
- A few dependencies (docker, requests, pyyaml) have been updated to support Python 3.12, which threw errors given our previous pinned versions. We should scrutinize CI failures as API changes in these could cause issues.
- I've removed an extraneous workflow (seems like it was for generating a token on-demand) that didn't seem to be called anywhere.

NOTE: setup.py still exists in this PR because the in-repo workflows need it, but it can be removed once this merges.


komish:
A note to reviewers: To see the effects of these changes, you mostly want to look at the CI PRs themselves, where you should see the pyproject.toml get picked up over the setup.py script. E.g.

https://github.com/openshift-helm-charts/sandbox/actions/runs/6883588742/job/18724448113#step:4:192

.github/workflows/mercury_bot.yml
```diff
@@ -14,16 +14,18 @@ jobs:
...
       - name: Set up Python 3.x Part 2
         run: |
           # set up python
           python3 -m venv ve1
-          cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
-          cd scripts && ../ve1/bin/python3 setup.py install && cd ..
+          cd scripts
+          ../ve1/bin/pip3 install -r requirements.txt
+          ../ve1/bin/pip3 install .
+          cd ..
```
mgoerens, line 28: Note that this is actually not needed as the current directory is not persistent across steps.

mgoerens:
> It seems setup.py-based installations are deprecated

@komish Just curious, could you please expand on this a little bit ? And share if you found any resources that have made you come to this conclusion ?

komish:
> > It seems setup.py-based installations are deprecated
> 
> @komish Just curious, could you please expand on this a little bit ? And share if you found any resources that have made you come to this conclusion ?

@mgoerens Sure thing. Here's how that particular "rabbit hole" unfolded. When using Python 3.12, I got this error trying to install our module:

```
 ❱ python setup.py install
Traceback (most recent call last):
  File "/path/to/openshift-helm-charts/development/scripts/setup.py", line 1, in <module>
    import setuptools
ModuleNotFoundError: No module named 'setuptools'
(venv.tools) 
```

That struck me as odd, but just tells me that my virtual environment didn't have setuptools available (which was previously installed for us at venv creation IIRC). Fine, I could install it - but I also came across this link (I thought it was through some stack trace at the command-line, but now I can't reproduce it) - https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html.

The PyPA documentation also describes setup.py based installations as legacy: https://pip.pypa.io/en/stable/reference/build-system/

And finally, The Setuptools documentation describes how to replace it with a pyproject.toml: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

According to that doc, it would seem we could also move some of our setup.cfg to that file, but I'll don't see it necessary quite yet. I can definitely look to move that direction in the future.

---

That's how I ended up looking at what to do to replace the setup.py script. I also came across a ton of other tools in the same space (E.g. poetry), but opted to keep things close to what we had for now. At least until we have enough problems  in our dev environments with it to justify the move to new tooling.

.github/workflows/mercury_bot.yml
komish: If you mean the `cd` components, that's probably fair. I just kept it consistent with what's in other files. We recently modified the way these ran in a PR, but I don't remember if we split things into multiple lines or put things all in one line. Either way, I picked one and aligned it across all files that did this.

In the future, I think I'd like to replace this with the `make venv.tools` target, but that involves changing a bunch of paths to the invocations. It was more than I wanted to do in this PR.

I can back this change out if it's more comfortable. It doesn't impact the way this PR should work if I back it out.

mgoerens: I meant the `cd ../` at the end of the step. It doesn't persist to the next step.

I'm all for consistency across files and steps :+1: Let's leave it like that for now, it also doesn't do any harm.

mgoerens (approved):
/lgtm

Issue #300, opened 2023-11-16
Review metrics collection bug post release


komish:
Observed here in a recent certification. Issues with this have been addressed in recent PRs, but I haven’t looked into this new issue yet. https://github.com/openshift-helm-charts/charts/actions/runs/6897195931/job/18764943440?pr=1114

PR #301, opened 2023-11-17
Re-add function that metric execution calls

komish:
There's still a bug in the metrics collection script. Currently it's calling a function I removed. My apologies - I missed that it was being referenced by the metrics script here.

Closes https://github.com/openshift-helm-charts/development/issues/300


OchiengEd (approved):
/lgtm

PR #306, opened 2023-12-13
chore: use the same checkout configuration in test.yml as in build.yml

OchiengEd:
Credit to @nickboldt for this work. PR has been pulled from the `openshift-helm-charts/charts` repository.

GH action https://github.com/openshift-helm-charts/charts/actions/workflows/test.yml seems to be failing for the last few PR checks. This change might solve that problem?

However I'm not auth'd to run workflow code updates on this repo:

```
  # workflow only change but user not authorized
```

So a code owner will have to run this.

References:
https://github.com/openshift-helm-charts/charts/pull/1136

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,10 +34,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


komish (approved):
/lgtm

Issue #305, opened 2023-12-08
Identify root cause of environment setup issues In workflow test


komish:
Recent certification triggered a workflow_test run that failed. This workflow doesn’t technically run on chart certifications, but rather programmatically auto passes. A failure is unexpected. 

https://github.com/openshift-helm-charts/charts/actions/runs/7134584056/job/19429691212?pr=1133

komish:
This was recently addressed in https://github.com/openshift-helm-charts/development/pull/306 but will be reverted. 

The PR pull step doesn't actually need to change here. The root cause is that the a contributor submitted a PR to the production repository from an old base branch. Specifically, they submitted a PR sourced from our production repository at the 1.6.1 release. As of 1.6.2, we switched to python 3.10, which also required some changes to our dependency versions.

What happens is that our python environment within GHA (using Python 3.10) is trying to install the dependencies at the wrong versions, and that's what throws the error.

The "Workflow Test" test is really only used for version releases in our Charts/Stage/Sandbox repositories, and so it likely makes more sense for us to identify a better path to run this than on every PR (e.g. maybe we change a cookie file containing the new version and trigger on that). For now, I'm reverting the change.

komish:
Closing due to staleness. At this point, many things have changed in how our pipelines work, and we don't have the original context due to artifact expiration on GitHub Actions. If this crops up again, we'll resolve this.

PR #307, opened 2023-12-18
Revert 'chore: use the same checkout configuration in test.yml as in build.yml (#306)'

komish:
This reverts commit 06f0ba1df4bea2cc2662c0c93920c1474be47eb1.

See reasoning here: https://github.com/openshift-helm-charts/development/issues/305#issuecomment-1861431769


komish:
This'll need to be force merged because the token removal this reverts was needed to get through the workflow test.

OchiengEd (approved):
/lgtm

PR #308, opened 2023-12-19
Add contributor check gate

komish:
Borrows from our existing user check scripts to generate a smaller script that we can throw into various workflows as needed.

Some of our workflows fail for certification attempts even though they aren't associated with certification. In this cases, it makes sense to avoid having them run, but we still need them to run on pull requests. This PR introduces a workflow that gates those PRs so they don't run if the user isn't a repository owner. This gives the green check for partners/chart contributors (because the dependent workflow skips), but still allows us to use these workflows for their original use cases.

### DIFF
--- a/.github/workflows/check-contributor.yml
+++ b/.github/workflows/check-contributor.yml
@@ -1,0 +1,77 @@
+name: Check Contributor
+
+# Checks that the input user exists in the approvers section
+# of a given OWNERS file.
+#
+# Returns true/false at the is-repo-owner output.
+#
+# Can be configured to fail altogether for contexts where it makes sense,
+# but in cases where this needs to return a green check mark, it is the
+# the caller's responsibility to evaluate the is-repo-owner output to inform
+# whether to proceed with subsequent tasks.
+#
+# Intended for use with workflows triggered by pull_request_target (or similar)
+# events.
+
+on:
+  workflow_call:
+    inputs:
+      user:
+        type: string
+        required: true
+        description: 
+          the user to evaluate
+      fail-workflow-if-not:
+        type: boolean
+        required: false
+        default: false
+        description: |
+          fails this workflow if the contributor is not an owner, 
+          or the evaluation fails for any other reason
+    outputs:
+      is-repo-owner: # 'true' / 'false'
+        description: whether the input user is a repo owner
+        value: ${{ jobs.check-contributor.outputs.is-repo-owner }}
+jobs:
+  check-contributor:
+    outputs:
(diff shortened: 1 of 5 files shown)


.github/workflows/check-contributor.yml
```diff
@@ -0,0 +1,72 @@
...
+        value: ${{ jobs.check-contributor.outputs.is-repo-owner }}
+jobs:
+  check-contributor:
+    outputs:
+      is-repo-owner: ${{ steps.populate-output.outputs.is-repo-owner }}
+    name: Contributor is repo owner
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository base
+        uses: actions/checkout@v3
```
OchiengEd, line 43: Spaces between the individual steps might be good for legibility

scripts/src/owners/user_is_repo_owner.py
```diff
@@ -0,0 +1,77 @@
...
+
+try:
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Loader
+
+OWNERS_FILE = "OWNERS"
+
+
+def is_approver(username: str):
```
OchiengEd, line 24: I would also hint on the expected return type e.g:
```
def is_approver(username: str) -> bool:
    ...
```

scripts/src/owners/user_is_repo_owner.py
```diff
@@ -0,0 +1,77 @@
...
+    does not match our expectations.
+    """
+
+    with open(OWNERS_FILE, "r") as f:
+        data = f.read()
+        out = yaml.load(data, Loader=Loader)
+
+    if "approvers" not in out:
+        raise Exception(
+            'OWNERS file did not have the "approvers" key and it is required'
```
OchiengEd, line 37: This reads as a concatenation of two idea due to the presence of "and". I would simplify to say:
 `OWNERS file does not have the required "approvers" key`
 Alternatively,
 `OWNERS file does not container the "approvers" key which is required`

OchiengEd (approved):
Minor changes recommended but otherwise looks good

/lgtm

scripts/src/owners/user_is_repo_owner.py
komish: Fixed

scripts/src/owners/user_is_repo_owner.py
komish: Fixed

.github/workflows/check-contributor.yml
komish: Fixed

PR #315, opened 2024-01-30
Add tool-enforced import sorting

komish:
This PR extends the Ruff configuration to add a barebones, mostly isort-compatible import sorting linter. With the `ruff check --fix` flag, this sorts imports in-place in a similar style. Reasonable to expect: alphabetical per group, E.g. stdlib, local, external)


mgoerens (approved):
Neat !

PR #321, opened 2024-02-09
Rework owners_file.py and add custom exception

### LINKED ISSUE #317: Raise custom Exception in owners_file.py
[scripts/src/owners/owners_file.py ](https://github.com/openshift-helm-charts/development/blob/main/scripts/src/owners/owners_file.py) could be reworked to either:
1. Raise custom Exceptions whenever some data cannot be found in the OWNERS file. This requires also the caller functions to be modified to handle those Custom Exceptions. This is similar to the `status` return value in `get_owner_data_from_file` but using custom Exceptions instead.
2. Keep returning empty strings, but at least stop catching broad `Exceptions` in the code. This could be done by catching `KeyError` (potentially using `supress` context manager when nothing is done when catching the Exception) or by using `get()`.

mgoerens:
closes #317

### DIFF
--- a/scripts/src/owners/owners_file.py
+++ b/scripts/src/owners/owners_file.py
@@ -1,3 +1,4 @@
+import contextlib
 import os
 
 import yaml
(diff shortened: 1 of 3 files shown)


komish:
I would prefer changes of this size wait until after we finish rolling out the helm dependency bugfix.

mgoerens:
@komish now that the helm dependency bugfix is merged, we could look into getting this merged. I've rebased and fixed merge conflicts. Note that the work has been separated into two commits for better readability / reviewability.

scripts/src/owners/owners_file.py
```diff
@@ -8,70 +9,67 @@
...
-    return status, owner_content
+    success = True
+
+    try:
+        owner_content = get_owner_data_from_file(path)
+    except OwnersFileError as e:
+        print(f"Error getting OWNERS file data: {e}")
+        success = False
+
+    return success, owner_content
```
komish, line 26: Nit: Can we assert success instead of asserting failure?
```suggestion
    success = False

    try:
        owner_content = get_owner_data_from_file(path)
        success = True
    except OwnersFileError as e:
        print(f"Error getting OWNERS file data: {e}")

    return success, owner_content
```

scripts/src/owners/owners_file.py
```diff
@@ -8,70 +9,67 @@
...
-    vendor = ""
-    try:
-        vendor = owner_data["vendor"]["name"]
-    except Exception:
-        pass
-    return vendor
+    vendor_name = ""
+    with contextlib.suppress(KeyError):
+        vendor_name = owner_data["vendor"]["name"]
+    return vendor_name
```
komish, line 47: I was originally going to suggest this instead (to avoid importing `contextlib`):

```python
return owner_data.get("vendor", {}).get("name", "")
```

But, actually, my implementation (and yours) both suffer from a failure if owner_data is not a dictionary. AttributeError would get thrown for mine (calling the get method), and TypeError for yours (NoneType is not subscriptable).

Better to leave this as is?

---

[EDIT]: I'm seeing now that there are several places where this logic was being simplified, but we'll run into the problem cases because we can't assert the owner_data input's type.

komish (changes requested):
Couple of small things to consider. My apologies for the delay.

komish (approved):
/lgtm

PR #322, opened 2024-02-09
Remove unused function in 'signed_chart.py'

mgoerens:


komish:
Instead of removing it because it's unused, should we thinking about where the rest of its related functions are called and look to use it there?

As in, should we extending a workflow to assert what this function intends to assert?

komish:
@mgoerens I don't remember if we've discussed this one yet, so pinging you on this one to see where we landed re: the above comment ^

komish (approved):
/lgtm

mgoerens:
@komish I think your comment makes sense generally speaking. In that particular case I don't see a use for this function. Also it would need a rewrite.. I would not return `True` if we can't find a key in the owners file..

I will hit the merge button and remove this unused function

PR #326, opened 2024-02-14
Add more debug information when E2E tests fail

mgoerens:
When E2E tests fails due to a sandbox PR not running the expected Workflow, the PR number was until now not displayed, making it difficult to figure out which PR to look at in oder to get more information. This commit adds the PR number to the raised Exception message to make it visible in the workflow logs directly.


komish (approved):
/lgtm

Issue #296, opened 2023-11-10
Create extra job for extracting and validating all needed info from PR


mgoerens:
At several places in the build.yaml workflow, we extract and validate information from the PR, its files, etc. Typically we need the `category`, `organization`, `chart_name`,  `chart_version` at different times.

There are several occurrences of such redundant calls (e.g. to `get_modified_files`) at different stages of the workflow.

We should create a separate job that takes care of:
1. Extracting all information that is needed by later steps from the PR content.
2. Validate those info (e.g. valid semver version, match between directory structure and info in report.yaml, etc.)
3. Present those info to the other steps

For 3. instead of GitHub outputs, we could populate a file and pass it to the other jobs as a PR artifact (see https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow)

Such a file could look for example like this:

```yaml
pre-check:
    content:
        modified_files:
            - charts/partners/hashicorp/vault/1.13.0/report.yaml
            - ...
        category: partners
        organization: hashicorp # vendor
        version: 1.13.0
        chart_name: vault
        source:
            provided: False
        report:
            provided: True
            signed: False
        owners_file_modified: False
        web_catalog_only: True
    pr_infos:
        api_url:
        PR_number: 30145
```

komish:
@mgoerens 100% agree. Looking at the certified chart dependency project, I may need to implement a small portion of this to accomplish a few things required in that pipeline. I may continue to use GITHUB_OUTPUT for those cases, but we can transition that down to some kind of format that looks like the above down the line.

mgoerens:
See https://github.com/openshift-helm-charts/development/pull/314#discussion_r1471308712 for future work item

PR #332, opened 2024-03-07
Initialize the Submission data structure

mgoerens:
This is the first step towards solving #296. The code added by this commit is currently *not* put in use in the certification pipeline and is a partial duplicate of the "checkprcontent" package.

The Submission data structure can collect and run early validation checks on all information concerning a user's submission (i.e. a GitHub Pull Request). In its current state, given an api_url (link to a given PR), it can collect the list of modified files, and extract the category, organization, chart's name and version, as well as running basic checks, such as SemVer compatibility of the provided version.

Eventually, this data structure is meant to encapsulate more information concerning the user's submission, and be dumped to a json file (using JSONEncode / JSONDecoder, more on that in a later PR) to be passed around the certification pipeline jobs as an artifact. This early (and incomplete) version is intended to trigger discussion around the design choices and general direction taken to solve #296.

### DIFF
--- a/scripts/src/precheck/submission.py
+++ b/scripts/src/precheck/submission.py
@@ -1,0 +1,336 @@
+import os
+import re
+import semver
+
+from dataclasses import dataclass, field
+
+from checkprcontent import checkpr
+from tools import gitutils
+from reporegex import matchers
+
+xRateLimit = "X-RateLimit-Limit"
+xRateRemain = "X-RateLimit-Remaining"
+
+
+class SubmissionError(Exception):
+    """Root Exception for handling any error with the submission"""
+
+    pass
+
+
+class DuplicateChartError(SubmissionError):
+    """This Exception is to be raised when the user attempts to submit a PR with more than one chart"""
+
+    pass
+
+
+class VersionError(SubmissionError):
+    """This Exception is to be raised when the version of the chart is not semver compatible"""
+
+    pass
+
+
+@dataclass
+class Chart:
+    """Represents a Helm Chart
+
+    Once set, the category, organization, name and version of the chart cannot be modified.
+
+    """
+
+    category: str = None
+    organization: str = None
+    name: str = None
+    version: str = None
+
+    def register_chart_info(self, category, organization, name, version):
+        if (
+            (self.category and self.category != category)
+            or (self.organization and self.organization != organization)
+            or (self.name and self.name != name)
+            or (self.version and self.version != version)
+        ):
+            msg = "[ERROR] A PR must contain only one chart. Current PR includes files for multiple charts."
+            raise DuplicateChartError(msg)
+
+        if not semver.VersionInfo.isvalid(version):
+            msg = (
+                f"[ERROR] Helm chart version is not a valid semantic version: {version}"
+            )
+            raise VersionError(msg)
+
+        self.category = category
+        self.organization = organization
+        self.name = name
+        self.version = version
+
+    def get_owners_path(self):
+        return f"charts/{self.category}/{self.organization}/{self.name}/OWNERS"
+
+
+@dataclass
+class Report:
+    found: bool = False
+    signed: bool = False
+    path: str = None
+
+
+@dataclass
+class Source:
+    found: bool = False
+    path: str = None  # Path to the Chart.yaml
+
+
+@dataclass
+class Tarball:
+    found: bool = False
+    path: str = None
+    provenance: str = None
+
+
+@dataclass
+class Submission:
+    """Represents a GitHub PR, opened to either certify a new Helm chart or add / modify an OWNERS file.
+
+    A Submission can be instantiated either:
+    * by solely providing the URL of a given PR (represented by the api_url attribute). Upon
+    initialization (see __post_init__ method), the rest of the information is retrieved from the
+    GitHub API. This should typically occur once per pipeline run, at the start.
+    * by providing all class attributes. This is typically done by loading a JSON representation of
+    a Submission from a file, and should be done several times per pipeline runs, in later jobs.
+
+    """
+
+    api_url: str
+    modified_files: list[str] = None
+    chart: Chart = field(default_factory=lambda: Chart())
+    report: Report = field(default_factory=lambda: Report())
+    source: Source = field(default_factory=lambda: Source())
+    tarball: Tarball = field(default_factory=lambda: Tarball())
+    modified_owners: list[str] = field(default_factory=list)
+    modified_unknown: list[str] = field(default_factory=list)
+
+    def __post_init__(self):
+        """Complete the initialization of the Submission object.
+
+        Only retrieve PR information from the GitHub API if requiered, by checking for the presence
+        of a value for the modified_files attributes. This check allows to make the distinction
+        between the two aforementioned cases of initialization of a Submission object:
+        * If modified_files is not set, we're in the case of initializing a brand new Submission
+        and need to retrieve the rest of the information from the GitHub API.
+        * If a value is set for modified_files, that means we are loading an existing Submission
+        object from a file.
+
+        """
+        if not self.modified_files:
+            self.modified_files = []
+            self._get_modified_files()
+            self._parse_modified_files()
+
+    def _get_modified_files(self):
+        """Query the GitHub API in order to retrieve the list of files that are added / modified by
+        this PR"""
+        page_number = 1
+        max_page_size, page_size = 100, 100
+        files_api_url = re.sub(r"^https://api\.github\.com/", "", self.api_url)
+
+        while page_size == max_page_size:
+            files_api_query = (
+                f"{files_api_url}/files?per_page={page_size}&page={page_number}"
+            )
+            print(f"[INFO] Query files : {files_api_query}")
+
+            try:
+                r = gitutils.github_api(
+                    "get", files_api_query, os.environ.get("BOT_TOKEN")
+                )
+            except SystemExit as e:
+                raise SubmissionError(e)
+
+            files = r.json()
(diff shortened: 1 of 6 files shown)


scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+            self._parse_modified_files()
+
+    def _get_modified_files(self):
+        """Query the GitHub API in order to retrieve the list of files that are added / modified by this PR"""
+        page_number = 1
+        max_page_size, page_size = 100, 100
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f'Bearer {os.environ.get("BOT_TOKEN")}',
+        }
```
komish, line 116: This feels odd to see here. Is this a carryover from checkpr? Should this (and the code below where we define the endpoint and make the call itself) be using https://github.com/openshift-helm-charts/development/blob/main/scripts/src/tools/gitutils.py#L70-L82

scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+        * The tarball file is names incorrectly
+        * The PR contains additional files, not related to the Chart being submitted
+        * The user attempts to create the OWNERS file for its project.
+
+        """
+        pattern, reportpattern, tarballpattern = (
+            checkpr.get_file_match_compiled_patterns()
+        )
+        matches_found = 0
+        none_chart_files = {}
```
komish, line 163: NIT: Rename this (and all references).

```suggestion
        non_chart_files = {}
```

I know this is a carryover from other code. If you want to keep it the way it is for this PR to reduce the diff surface area, that's fine.

scripts/src/precheck/submission_test.py
```diff
@@ -0,0 +1,165 @@
+import pytest
+import responses
+import sys
+
+sys.path.append(".")
```
komish, line 5: Nit: I'm assuming we're doing this here so that we can reference submission in the following line, but this is a Python-ism that bugs me a bit. I'm reading this as "add the files in `.` as import names".

Normally, you would do `from precheck import submission`, and then reference the classes as `submission.Submission`, right?

Shouldn't we have access to that without manipulating the default import directories?

For what it's worth, I do recognize that we have this in other source files, which is what makes this a nit. Feel free to disregard.

scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+    report: Report = field(default_factory=lambda: Report())
+    source: Source = field(default_factory=lambda: Source())
+    tarball: Tarball = field(default_factory=lambda: Tarball())
+
+    def __post_init__(self):
+        """Complete the initialization of the Submission object.
+
+        modified_files should be gatered and validated at the beginning of the pipeline, as part
+        of the "pre-check" job. Hence this isn't done when a Submission object is loaded from a
+        JSON file aka when modified_files is not empty.
```
komish, line 102:
```suggestion
        modified_files should be gated and validated at the beginning of the pipeline, as part
        of the "pre-check" job. Hence this isn't done when a Submission object is loaded from a
        JSON file aka when modified_files is not empty.
```

Also, I may be misreading, but I'm not understanding the context of this comment.

scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+@dataclass
+class Tarball:
+    found: bool = False
+    path: str = None
+    provenance: str = None
+
+
+@dataclass
```
komish, line 88: Because this is a dataclass, I would understand that all of the class variables defined would be settable by default (in the generated init function). When I'm running this locally, it feels like I generally only want to pass in api_url, and the rest of the fields _should_ be derived.

Do I have a bad read on the application of this class, or should the rest of these class variables not be settable?

komish (changes requested):
Great work here. Overall design looks good, and seems effective. Couple of thoughts in-lined.

scripts/src/precheck/submission.py
mgoerens: Good question.

TLDR: A constructor that accepts *all* attributes is needed for the JSONDecoder to function. Specifically [here](https://github.com/openshift-helm-charts/development/pull/332/files/5a55ce326e4cde2a43d0f5ad634d7604afaea6c9#diff-ded7b02fcbdf810b9bf20e8f306d88f09f94202e4b3cf8a49149b55ab05bfcb4R39).

I envision two cases where we are going to instantiate a `Submission` object:
1. In the "pre-check" job (we can revisit the name of this job BTW), we are given an api_url, and the rest is derived. At the end of this process, we dump the Submission object as JSON to an [artifact file](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow).
2. In following jobs, we want to access some of the Submission information, such as the path of the report, or the category. Either we do that directly in the jobs shell (`run`) using `jq`, or we have a python script that loads the artifact file. The later can be easily achieved with the `JSONDecoder` (Note that this is needed due to the nested objects: `Chart`, `Report`, `Source` and `Tarball`).

In order to know if we're dealing with 1. or 2., we check in `__post_init__` if we were provided with a list of `modified_files`. This way we don't populate the class again from scratch and override the given input data (e.g. artifact file) - though of course it should find the exact same info.

That's what I was trying to explain in [this comment](https://github.com/openshift-helm-charts/development/pull/332/files/5a55ce326e4cde2a43d0f5ad634d7604afaea6c9#diff-6cdd6c8509751e3be6c04f9c38d4c43c16760aab0318c9ed6cbad88bc9b89300R100-R102) that you didn't understand. I don't blame you, I haven't done an amazing job at explaining :) I will rework the comment.

scripts/src/precheck/submission.py
mgoerens: Intention was to be "gathered". I will provide more information, that was clear to me at the time of writing but as a new reader (and even to me now) it's a confusing and incomplete formulation.

scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def register_chart_info(self, category, organization, name, version):
+        if (
+            (self.category and self.category != category)
+            or (self.organization and self.organization != organization)
+            or (self.name and self.name != name)
+            or (self.version and self.version != version)
+        ):
+            msg = "[ERROR] A PR must contain only one chart. Current PR includes files for multiple charts."
+            print(msg)
```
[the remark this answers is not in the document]
mgoerens: I fully agree. There are carried over from the `checkprcontent` library. There they usually look like this:
```
                msg = "[ERROR] A PR must contain only one chart. Current PR includes files for multiple charts."
                print(msg)
                gitutils.add_output("pr-content-error-message", msg)
                exit(1)
```

Eventually I hope to deprecate `checkpr` in favor of  `Submission` (though they might have to be co-living for a period of time).

What I intend to do is to catch the Exception in the script that will call this library, and handle `print` and `gitutils.add_output` there.

So it's a left-over, kind of a reminder for my future self. I will remove the print statements as I agree it's not needed (also not to remind my future self).

scripts/src/precheck/submission.py
```diff
@@ -0,0 +1,220 @@
...
+        max_page_size, page_size = 100, 100
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f'Bearer {os.environ.get("BOT_TOKEN")}',
+        }
+        files_api_url = f"{self.api_url}/files"
+
+        while page_size == max_page_size:
+            files_api_query = f"{files_api_url}?per_page={page_size}&page={page_number}"
+            print(f"[INFO] Query files : {files_api_query}")
```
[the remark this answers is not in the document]
mgoerens: I will keep this in mind, thanks

scripts/src/precheck/submission.py
mgoerens: Yes, it's another carryover. I will look into using `gitutils`.

komish: If it's a carry over, you can leave it as is. I don't want you to have to change a bunch of things related to copied code if it works as is and doesn't impact what you're trying to implement.

scripts/src/precheck/submission.py
mgoerens: Good catch, I didn't even realized this typo in the original code ! I will change it.

FYI I'm not happy with the current implementation, especially of the `_parse_modified_files` method, but I needed something that works so I stole the code from checkpr. I would like to revisit it before getting this merged ideally (more to come in a further commit).

My intention was to discuss the general design before spending too much time refining the implementation.

In other word, I don't think the "diff surface area" is going to be small anyway and don't mind taking the time to rewrite part of the existing script, rather than trying to use the exact same code as much as possible.

komish: > FYI I'm not happy with the current implementation, especially of the _parse_modified_files method, but I needed something that works so I stole the code from checkpr. I would like to revisit it before getting this merged ideally (more to come in a further commit).

Totally understand. 100% agree. It could use a clarity rewrite, but it works so I haven't prioritized it.

mgoerens:
@komish I removed the DRAFT status as I believe this is now mature enough to be reviewed again and eventually merged with the current set of capabilities. I intend to continue iterate over this work, but this PR is already long enough, and I'd rather create smaller PR in the future.

This PR has seen significant changes since you reviewed it first:
- Most impacting change is that I account for another type of "Submission", when a user submits only the OWNERS file.
- Consequently, I added the `is_valid_certification_submission` and `is_valid_owners_submission` methods.
- Reworked tests

komish (approved):
/lgtm

PR #343, opened 2024-04-11
Ping maintainers if the release job fails

### LINKED ISSUE #341: Pipeline is silently failing when index update fails
See: https://github.com/openshift-helm-charts/charts/actions/runs/8518766581/job/23331558923?pr=1359

Reason for the fail is tracked [in another issue](https://github.com/openshift-helm-charts/development/issues/342).

In this pipeline, the "Release Chart" job failed. As a result, the index was never updated and the chart never actually released. Also, this allowed the partner to update the chart [in a new PR](https://github.com/openshift-helm-charts/charts/pull/1368/), although that shouldn't be allowed (see the index check: https://github.com/openshift-helm-charts/charts/blob/main/scripts/src/checkprcontent/checkpr.py#L224)

Problem is that this issue was never communicated to the user nor to us maintainers. While the pipeline do fails, the PR is already merged at this stage (merge occurs at the end of the "chart-verifier" job, run previously), and no error appears. We should at least be informed.

I believe the correct solution is to add a comment on the PR that pings maintainers in case of a failure in the "Release Chart" job.

mgoerens:
Add a GitHub comment if an error occurs during the update of the helm index that pings the maintainers.

fix #341

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -533,6 +533,23 @@ jobs:
             https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} \
             HEAD:refs/heads/${INDEX_BRANCH}
 
+      - name: Add a GitHub comment if release has failed
+        uses: actions/github-script@v7
+        if: ${{ failure() && env.GITHUB_REPOSITORY != 'openshift-helm-charts/sandbox' }}
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: `### Release job failed
+
+              An error occured while updating the Helm repository index.
+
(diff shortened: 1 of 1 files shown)


mgoerens:
FYI failure in CI were caused by either:
- the workflow did not run
- in the release job, GitHub release not found (404)

In both cases, I tend to think those issues are caused by GitHub slowness, and not by the changes in this PR. Will retry in a second when I push an updated commit.

komish:
@mgoerens LGTM. Good catch.

Two non-blocking comments:

The only thing I'd mention here is that [this approach](https://github.com/openshift-helm-charts/development/blob/main/.github/workflows/check-locks-on-owners-submission.yml#L155-L173), which I've found only recently, feels a bit simpler to comment on a PR. I like github-script, but in some cases it requires more error handling than I really care to deal with. Here, it looks fine to me.

Secondly, a failure to index a merged chart might warrant a more direct ping than a GH comment. We may consider Slack comms in the future.

Non-blocking, so nothing to change here.

Thanks!

komish (approved):
/lgtm

PR #344, opened 2024-04-16
Add check for web_catalog_only to Submission

mgoerens:
This commits ports the "check_web_catalog_only" from the checkpr package over to the Submission data structure.

PR #345, opened 2024-04-17
Inform caller if web_catatog_only key is missing

mgoerens:
Add a "success" return value to get_web_catalog function to inform the caller if the file (OWNERS or report.yaml) doesn't contain such a key.

While this is currently not used by callers, this is a pre-requisite for #344 to exit draft status

### DIFF
--- a/scripts/src/owners/owners_file.py
+++ b/scripts/src/owners/owners_file.py
@@ -13,6 +13,10 @@ class OwnersFileError(Exception):
     pass
 
 
+class ConfigKeyMissing(Exception):
+    pass
+
+
 def get_owner_data(category, organization, chart):
     path = os.path.join("charts", category, organization, chart, "OWNERS")
     success = False
@@ -61,7 +65,28 @@ def get_chart(owner_data):
     return chart
 
 
-def get_web_catalog_only(owner_data):
+def get_web_catalog_only(owner_data, raise_if_missing=False):
+    """Check the delivery method set in the OWNERS file data
+
+    Args:
+        owner_data (dict): Content of the OWNERS file. Typically this is the return value of the
+            get_owner_data or get_owner_data_from_file function.
+        raise_if_missing (bool, optional): Whether to raise an Exception if the delivery method is
+            not set in the OWNERS data. If set to False, the function returns False.
+
+    Raises:
+        ConfigKeyMissing: if the key is not found in OWNERS and raise_if_missing is set to True
+
+    """
+    if (
+        "web_catalog_only" not in owner_data
+        and "providerDelivery" not in owner_data
+        and raise_if_missing
+    ):
+        raise ConfigKeyMissing(
+            "Neither web_catalog_only nor providerDelivery keys were set"
+        )
+
     return owner_data.get("web_catalog_only", False) or owner_data.get(
         "providerDelivery", False
     )
--- a/scripts/src/report/verifier_report.py
+++ b/scripts/src/report/verifier_report.py
@@ -43,6 +43,10 @@ SUPPORTED_VERSIONS_ANNOTATION = "charts.openshift.io/supportedOpenShiftVersions"
 KUBE_VERSION_ATTRIBUTE = "kubeVersion"
 
 
+class ConfigKeyMissing(Exception):
+    pass
+
+
 def get_report_data(report_path):
     """Load and returns the report data contained in report.yaml
 
@@ -105,20 +109,41 @@ def get_profile_version(report_data):
     return profile_version
 
 
-def get_web_catalog_only(report_data):
+def get_web_catalog_only(report_data, raise_if_missing=False):
+    """Check the delivery method set in the report data.
+
+    Args:
+        report_data (dict): Content of the report file. Typically this is the return value of the
+            get_report_data function.
+        raise_if_missing (bool, optional): Whether to raise an Exception if the delivery method is
+            not set in the report data. If set to False, the function returns False.
+
+    Raises:
+        ConfigKeyMissing: if the key is not found in OWNERS and raise_if_missing is set to True
+
+    """
+    keyFound = False
     web_catalog_only = False
     try:
         if "webCatalogOnly" in report_data["metadata"]["tool"]:
             web_catalog_only = report_data["metadata"]["tool"]["webCatalogOnly"]
+            keyFound = True
         if "providerControlledDelivery" in report_data["metadata"]["tool"]:
             web_catalog_only = report_data["metadata"]["tool"][
                 "providerControlledDelivery"
             ]
+            keyFound = True
     except Exception as err:
         print(
             f"Exception getting webCatalogOnly/providerControlledDelivery {err=}, {type(err)=}"
         )
         pass
+
+    if not keyFound and raise_if_missing:
+        raise ConfigKeyMissing(
+            "Neither webCatalogOnly nor providerControlledDelivery keys were set"
+        )
+
     return web_catalog_only
 
 


scripts/src/owners/owners_file.py
```diff
@@ -62,9 +62,19 @@ def get_chart(owner_data):
...
+
+    with contextlib.suppress(KeyError):
+        owners_web_catalog_only = owner_data["web_catalog_only"]
+        success = True
+
+    with contextlib.suppress(KeyError):
+        owners_providerDelivery = owner_data["providerDelivery"]
+        success = True
+
+    return success, owners_web_catalog_only or owners_providerDelivery
```
komish, line 77: Bit of a preference, but I'd like to see the result order inverted such that the answer to the function call "get_web_catalog_only" is the first returned parameter, and whether or not the key was actually found, or the returned value was a fallback false is the second. Similar to what you'd see in Go: 

```go
if theValue, exists := mymap["foo"]; exists {
    ...
}
```

komish: Also, `success` isn't as meaningful as I'd like. Better to have `keyFound` or something that tells us clearly what was returned.

komish (changes requested):
One small change.

That said, I'm having trouble understanding why we would need to know if the key existed or not. I reviewed the draft PR this blocks, and It wasn't quite clear to me what decisions we're making where we actually care if the value *exists* vs. the meaning of the value, which we already get back from this function.

mgoerens:
> That said, I'm having trouble understanding why we would need to know if the key existed or not. I reviewed the draft PR this blocks, and It wasn't quite clear to me what decisions we're making where we actually care if the value _exists_ vs. the meaning of the value, which we already get back from this function.

This is were where I check for the value and its existence in the OWNERS: https://github.com/openshift-helm-charts/development/pull/344/commits/d67c95850a85e020e566d489747d7dc3a44d5733#diff-6cdd6c8509751e3be6c04f9c38d4c43c16760aab0318c9ed6cbad88bc9b89300R341-R345

And similarly, a few lines later, in the report: https://github.com/openshift-helm-charts/development/pull/344/commits/d67c95850a85e020e566d489747d7dc3a44d5733#diff-6cdd6c8509751e3be6c04f9c38d4c43c16760aab0318c9ed6cbad88bc9b89300R354-R360

The problem is that the function previously defaults to returning `False`. In the event the key is not present in OWNERS, I want to return an error. 

I made the assumption that this key is mandatory, both in OWNERS and in report.

komish:
@mgoerens For the purposes of certification, however, a value of "False" (whether the key is found and set to false, key is found and set to null, or key is not found) is still a valid value, right?

The linked logic seems to intend to parse a given submission, and populate the CI environment with useful representations of the project's context. E.g. "This is a webCatalogOnly project; This is a Partner project". For that purpose, the existing logic makes sense. I don't know if we want to raise errors for OWNERS file content on chart submissions when the format of the OWNERS file could be validated in other contexts (e.g. when submitted).

Does that make sense? Is your intention to use this Submission class for every PR, regardless of its purpose?

mgoerens:
@komish Sorry, took some time before coming back to this.

> Is your intention to use this Submission class for every PR, regardless of its purpose?

I believe we have 2 types of PRs: ones that add / modify an OWNERS file, and one that adds a chart. I try to design the Submission class in a way that is can be initialized in both cases (calling `my_submission = Submission(api_url=...)` then uses methods `is_valid_owners_submission()` and `is_valid_certification_submission()` to check.

However this particular method is only meant to be called on "Certification" type of Submissions.

Could be designed differently, I'm thinking of using Subclasses. Having `is_valid_...` type of methods also makes it easier to write unit tests BTW. But I'm digressing.

> I don't know if we want to raise errors for OWNERS file content on chart submissions when the format of the OWNERS file could be validated in other contexts (e.g. when submitted).

It should definitely be validated when submitted. But still... what if it's somehow absent ? There isn't really a good default value for web_catalog_only. It feels odd to me to default to False if absent. Generally speaking, what to do with a Chart submission if the OWNERS is of invalid format ? I'd rather error the certification.

Note that I will address your comments reg. code once we settle the discussion here. I don't want a bunch of linked PRs to clutter the GitHub UI ... you know ...

komish:
Hey @mgoerens! I'm just gettin back to this, sorry for the delay - let's finish this up so we can get it merged in.

> It should definitely be validated when submitted. But still... what if it's somehow absent ? There isn't really a good default value for web_catalog_only.

I suppose this is a valid point. To a certain extent, for simpler logic like these lookups, the library is trying to "tell you the information you want when calling the function", which is - if it's missing, then it is not explicitly set to "true" and therefore should return "false" - and I tend to lean this way.

With that said, I think we could do a couple of things here to get the best of both. LMK how you feel about these options

1) Optionally `Raise` an exception if the key is missing, e.g.

```python
# pseudocode
def get_web_catalog_only(owner_data, raise_if_missing=False):
    success = False
    
    #
    # ... code omitted for brevity ...
    #

    if not success and raise_if_missing:
        raise(ConfigKeyMissing("Neither web_catalog_only nor providerDelivery keys were set"))

    return owners_web_catalog_only or owners_providerDelivery
```

Doesn't change the return, and should allow you to catch `ConfigKeyMissing` (which is a made up exception not defined in the snippet) in cases where you want to return an error. You don't have to pass the ConfigKeyMissing as the final exception, but catching it should allow you to return whatever you want in cases where you're validating.

2) Use the `get` method on maps to resolve the "unset" nature of the `web_catalog_only` and/or `providerDelivery` keys. E.g.

```python
(owners_file.get('web_catalog_only', 'unset') == 'unset') and (owners_file.get('providerDelivery', 'unset') == 'unset')
```

mgoerens:
@komish

I like the idea of using `raise_if_missing` a lot. That allows for the best of both world indeed.

I'm not sure I follow the second idea though. What does this solve ?

komish:
The second option just pushes the responsibility of checking whether or not the key is set to the caller of the original function _at call time_. It's much longer, much more verbose, but leaves the original functions intact and accomplishes the same thing.

scripts/src/owners/owners_file.py
```diff
@@ -61,7 +65,28 @@ def get_chart(owner_data):
...
+        raise_if_missing (bool, optional): Wether to raise an Exception if the delivery method is
+            not set in the OWNERS data. If set to False, the function returns False.
+
+    Raises:
+        ConfigKeyMissing: if the key is not found in OWNERS and raise_if_missing is set to True
+
+    """
+    if (
+        "web_catalog_only" not in owner_data
+        and "providerDelivery" not in owner_data
```
komish, line 83: Should these be `( ... )` grouped? Or does it matter?

scripts/src/owners/owners_file.py
```diff
@@ -61,7 +65,28 @@ def get_chart(owner_data):
...
 
 
-def get_web_catalog_only(owner_data):
+def get_web_catalog_only(owner_data, raise_if_missing=False):
+    """Check the delivery method set in the OWNERS file data
+
+    Args:
+        owner_data (dict): Content of the OWNERS file. Typically this is the return value of the
+            get_owner_data or get_owner_data_from_file function.
+        raise_if_missing (bool, optional): Wether to raise an Exception if the delivery method is
```
komish, line 74: NIT: can be fixed in a future PR if you want.

```suggestion
        raise_if_missing (bool, optional): Whether to raise an Exception if the delivery method is
```

komish (approved):
/lgtm

Small comments left, up to you if you want to interrupt CI to fix them. I think both of them are negligible.

scripts/src/owners/owners_file.py
mgoerens: It would be the same logically to have `A & B & C` or `(A & B) & C`, and IMO would only clutter the code.

komish (approved):
/lgtm

PR #347, opened 2024-04-23
Ports checks if a chart is in index; and if a release tag already exists

mgoerens:
This is a port of the two following checks, performed at the end of the
checkpr script, to the Chart data structure:
* Check that the Chart in this version is not already present in the
  index
* Check that the release tag for this chart doesn't already exists on a
  given GitHub repository

Reviewers, please see https://github.com/openshift-helm-charts/development/blob/main/scripts/src/checkprcontent/checkpr.py#L224-L276 for the original code (still in use at the moment).

### DIFF
--- a/scripts/src/precheck/submission.py
+++ b/scripts/src/precheck/submission.py
@@ -1,6 +1,13 @@
 import os
 import re
+import requests
 import semver
+import yaml
+
+try:
+    from yaml import CLoader as Loader
+except ImportError:
+    from yaml import Loader
 
 from dataclasses import dataclass, field
 
@@ -33,6 +40,14 @@ class VersionError(SubmissionError):
 
 
 class WebCatalogOnlyError(SubmissionError):
+    pass
+
+
+class HelmIndexError(SubmissionError):
+    pass
+
+
+class ReleaseTagError(SubmissionError):
     pass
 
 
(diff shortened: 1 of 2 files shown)


scripts/src/precheck/submission.py
```diff
@@ -73,6 +88,53 @@ def register_chart_info(self, category, organization, name, version):
...
+        Raise:
+            HelmIndexError if:
+            * The provided index is malformed
+            * The Chart is already present in the index
+
+        """
+        try:
+            chart_entry = index["entries"].get(self.name, [])
+        except KeyError as e:
+            raise HelmIndexError(f"Malformed index {index}") from e
```
komish, line 109: If I'm reading this right, we'd never see a key error here because you're returning a default value.

```python3
>>> x = {}.get("foo", [])
>>> x
[]
```

scripts/src/precheck/submission.py
```diff
@@ -439,3 +501,17 @@ def get_file_type(file_path):
...
+def download_index_data(repository, branch="gh_pages"):
+    """Download the helm repository index"""
+    r = requests.get(
+        f"https://raw.githubusercontent.com/{repository}/{branch}/index.yaml"
+    )
+
+    if r.status_code == 200:
+        data = yaml.load(r.text, Loader=Loader)
+    else:
+        data = {"apiVersion": "v1", "entries": {}}
```
komish, line 515: I know this is coming from the original code, so I'm not going to worry too much about this now - but if the GitHub API returns a non-200 OK, do we want to fall back to this? For example, wouldn't this produce a "chart version is available because it doesn't exist in the mapping" response in the code you're porting here?

komish (commented):
One specific comment inlined.

Generally - I did notice that a lot of the logging that was happening in the original code is now missing. Do you have something in mind for replacing them?

scripts/src/precheck/submission.py
mgoerens: Yes it would, and I don't think this is the right approach.

We could instead raise an Exception in case the GitHub API is not available / returns a non-200 status code. WDYT ?

mgoerens:
@komish Reg. logging, yes I have a plan. I want to have a `main()` function in the `precheck` package, that instantiates the Submission and run the required checks. Exceptions should be caught and logged there.

I will also take some time to compare what is being logged in the legacy code vs what is being raised in the new code. Should be relatively close, but I'll double check if I missed anything important before the switch.

scripts/src/precheck/submission.py
komish: That sounds reasonable enough to me.

Up to you if you want to do that here, or do that in the next PR. If you decide to add it to the next PR, it should fast-follow this work.

mgoerens: I decided to simply return an empty dict, and leave the handling to the caller. Let me know if that's fine with you.

(There's currently no caller for this function)

komish (approved):
/lgtm

PR #349, opened 2024-05-15
New chart fixture with plus sign in version

jsm84:
Summary:
* Use redhat-cryostat chart as a fixture to test plus signs in version


komish (approved):
/lgtm

PR #350, opened 2024-05-20
Added new test scenario plus-in-version to HC-17

jsm84:
Summary:
* Added a new test scenario (HC-17-002) which tests a chart with a plus sign in the version
* Includes a pair of reports (`partner` and `redhat` vendor types) referencing a helm chart.tgz with a plus in the version


komish (approved):
/lgtm

PR #351, opened 2024-05-21
Rename crysostat fixture & rm whitespace in provider name

jsm84:
Summary:
* The new HC-17-002 test was failing due to two reasons:
  1. The chart contained the `redhat-` prefix (partner submission)
  2. The provider name `Red Hat` was incorrect (should be `RedHat`)
* This fixes both issues with the cryostat chart itself. New reports will follow in a separate PR.


komish (approved):
/lgtm

PR #352, opened 2024-05-23
Cleanup unused report fixtures for HC-16

jsm84:
Summary:
* There was a duplicate `dash-in-version` pair of report fixtures under `tests/data/HC-16/` that aren't referenced by the HC-16 test.
* These appear to be duplicates of the HC-17 fixtures under `tests/data/HC-17/`, so they can be removed.


komish (approved):
/lgtm

PR #353, opened 2024-06-03
Use Github context in tasks conditions

mgoerens:
We don't want to be pinged from sandbox (see this PR for instance https://github.com/openshift-helm-charts/sandbox/pull/92390). Hence fixing broken conditioning


komish (approved):
/lgtm

Weird, I feel like the env context should also have GITHUB_REPOSITORY... but this looks better anyway.

PR #356, opened 2024-06-11
Wrap owners file content access function in try except

komish:
The return values for `owners_file.get_owner_data_from_file` was [changed recently](https://github.com/openshift-helm-charts/development/pull/321) such that it no longer returns a boolean reflecting whether accessing the OWNERS file was successful, instead favoring a Raised exception. 

This PR updates the chart lock generation logic. See https://github.com/openshift-helm-charts/charts/actions/runs/9470017334


jsm84 (approved):
LGTM

PR #359, opened 2024-06-20
Update script dependencies and update fn calls to match

komish:
### Context: 

I'm looking to enable dependabot for this repository soon. Before that happens, I ran [dependabot in a fork](https://github.com/komish/charts-development/pull/4) and [ran the E2E testing](https://github.com/komish/charts-development/actions/runs/9572122603) to see how it behaved. It's look all green after a couple of signature changes (captured here), so I think this should be in good shape.

Please keep an eye out for anything strange that we think may be a result of these dependency changes. Several packages iterated major versions.

Dependabot enablement will happen later, along with syft/grype.

### Summary:

- Update dependencies as dependabot would
- Remove pytest-forked and pytest-xdist, which we aren't using (since the move to behave) and were pulling in a dependency we don't need (py).
- Update method calls that were renamed between a package version update.


jsm84 (approved):
@komish LGTM...but as a side question, I suppose only `semver.isvalid` needed changing in our code?

komish:
> @komish LGTM...but as a side question, I suppose only `semver.isvalid` needed changing in our code?

Correct. Did I change it somewhere unexpected or miss a spot?

PR #360, opened 2024-06-27
Add dependabot configuration for core scripts

komish:
- Repository configuration will be updated to enable dependabot after this merges.
- Release info updated so that this configuration doesn't end up in the various other downstream repos.


jsm84 (approved):
lgtm

PR #362, opened 2024-07-02
Run Python unit tests in GitHub action

mgoerens:


komish:
@mgoerens please add this file to release_info.json so that it does not get copied to other repositories. Should be similar to the change I made here https://github.com/openshift-helm-charts/development/pull/360/files#diff-b474dc8e6d039d028f37776c76a3ac558d243af475173fd9d4ca8931cd960dee. In effect, just adding the file to the mapping(s) without changing the release version should not trigger release-related workflows that also react to modifications of this file.

Otherwise LGTM.

komish (approved):
/lgtm

PR #363, opened 2024-07-09
Add redhat prefix check

mgoerens:
Ported from:
https://github.com/openshift-helm-charts/development/blob/919146bc199e6366c40cf10d533d4669bd3def70/scripts/src/checkprcontent/checkpr.py#L211-L222


komish (approved):
/lgtm

PR #366, opened 2024-07-17
Ensure service account token generation for workflows

### LINKED ISSUE #365: Service Account tokens are no longer generated, breaking on-PR namespace generation
The `LegacyServiceAccountTokenNoAutoGeneration` feature gate has been removed as of OpenShift 4.15 (we're on 4.16). Our scripts that create namespaces on-demand and on-PR now need to also create a token secret so that the in-cluster token service can generate a token for use in the workflow.

These scripts need to be updated to accommodate this change:

1)  https://github.com/openshift-helm-charts/development/blob/main/scripts/src/saforcertadmin/create_sa.sh

This needs to be updated to create a token secret.

2)  https://github.com/openshift-helm-charts/development/blob/main/scripts/src/saforcharttesting/saforcharttesting.py

This script needs to be updated to create a token secret, and then read the value from that secret.

komish:
Fixes #365 

For the saforcertadmin scripts, we would previously detect the service account secret and read that for the token value. Since we now know the name of the secret, the logic becomes simpler.

For the saforcharttesting scripts, we just add an additional resource to be created. I left the parsing logic intact as it seems to work in my testing, and remains backwards compatible (not that we really need that here).

### DIFF
--- a/scripts/src/saforcertadmin/create_sa.sh
+++ b/scripts/src/saforcertadmin/create_sa.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 user_name='rh-cert-user'
+token_secret='rh-cert-user-token'
 oc create sa $user_name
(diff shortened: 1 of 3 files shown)


jsm84 (approved):
LGTM

jsm84 (approved):
Sorry I missed this. LGTM

PR #368, opened 2024-07-18
Allow actions to use node16 temporarily

komish:
The previously implemented workaround saved us some time, but not enough. It looks like the core issue is that node20 is actually running for workflows that run are setup to use node16. In my testing, node20 exhibits the issue but node 16 and 18 do not, so I'm temporarily enabling node16 until we can get the tools installer to bump their version.

Also reverting the disabled cache.

https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


OchiengEd (approved):
/lgtm

PR #373, opened 2024-07-25
Extract temporary service account token by name directly

komish:
In https://github.com/openshift-helm-charts/development/pull/366, I modified the code that creates namespaces, RBAC resources, etc. used in certification workflows because OpenShift 4.16 (which is our current runtime platform) did not automatically populate tokens for the service accounts. In effect, I just manually created a secret so that the cluster would populate it with a token we could use.

Since then, I've noticed we had intermittent failures in nightly runs. 
https://github.com/openshift-helm-charts/sandbox/actions/runs/10060735219/job/27809079797#step:17:18

When we search for the service account token, we parse the service account in the temporary namespace to see if it has secrets populated. In certain (or all modern) cases, the token secret may not be listed in the service account's list of secrets, so the code would then use `oc describe` to list a token secret, then parse that using regular expressions to find secret name, then extract the value, and so on.

The failure we were seeing was because we would sometimes get a service account manifest back (in JSON format) with no `secrets` key. I don't know why, but that's neither here nor there. I had originally just wrapped this bit of code in a try/except KeyError block, but after reading it, I think decided to simplify it a bit.

---

This PR refactors this code to extract the token from the token secret directly because we _now_ know it's name (because we create it in this same code). 

In the previous iteration of this code (described above), the token secret would have been generated by Kubernetes with a randomized name. Given we didn't know that name we would use regular expressions to match any possible names, and then read data from it.

Now that we know the token secret we care about will have a fixed name, we can query it directly and extract the temporary token from that.

---

Aside: this script and library needs some refactor, but I didn't do that here/yet. I'm just addressing the intermittence to reduce invalid E2E failures we've been seeing in nightly runs. Refactoring this is a future task.

### DIFF
--- a/scripts/src/saforcharttesting/saforcharttesting.py
+++ b/scripts/src/saforcharttesting/saforcharttesting.py
@@ -2,7 +2,6 @@ import argparse
 import base64
 import json
 import os
-import re
 import subprocess
 import sys
 import tempfile
@@ -29,7 +28,7 @@ apiVersion: v1
 kind: Secret
 type: kubernetes.io/service-account-token
 metadata:
-  name: token-${name}
+  name: ${name}
   namespace: ${name}
   annotations:
     kubernetes.io/service-account.name: ${name}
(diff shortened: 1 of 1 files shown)


scripts/src/saforcharttesting/saforcharttesting.py
```diff
@@ -242,65 +241,44 @@ def delete_clusterrolebinding(name):
...
-                match = re.search(pattern, dstdout)
-                if match:
-                    token_name = match.group(1)
-                else:
-                    print("[ERROR] Token not found, Exiting")
-                    sys.exit(1)
-                secrets.append({"name": token_name})
-                secret_found = True
-                break
-        time.sleep(10)
```
acornett21, line 281: Do you still need/wait to sleep some? Or is this no longer important/needed because the secrete is being proactively created?

komish: The sleep was moved to the top of the function, and is only executed on retries.

acornett21 (approved):

mgoerens (approved):

PR #377, opened 2024-08-08
Standardize runner images to ubuntu-22.04

komish:
This replaces all ubuntu-* references that are not 22.04 to explicitly use 22.04 (which is the equivalent of ubuntu-latest at the time of this writing)

### DIFF
--- a/.github/workflows/check-contributor.yml
+++ b/.github/workflows/check-contributor.yml
@@ -37,7 +37,7 @@ jobs:
     outputs:
       is-repo-owner: ${{ steps.populate-output.outputs.is-repo-owner }}
     name: Contributor is repo owner
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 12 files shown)


mgoerens (approved):
LGTM, now I'm wondering about process to update to a newer version:

- Is there a way to be automatically notified when a newer version is available ? Note that `ubuntu-24-04` is [available](https://github.com/actions/runner-images/issues/9848) already, though it seems `ubuntu-latest` still points to 22.04.

- Can we configure something similar to dependabot for runner images ?

komish:
Indeed, I did notice ubuntu-24.04 is available but opted not to use it _yet_ given we weren't using 24.04 in any of our tasks.  Our dependabot configuration doesn't do workflows yet, but I'm also curious if enabling it would update runner images. I would guess it doesn't? But, I'm not sure. We can look into it.

PR #382, opened 2024-09-23
Remove unused GitHub output

mgoerens:
This GitHub output doesn't seem to be needed anymore as it's not used anywhere.


komish (approved):
I couldn't find any references to this either.

/lgtm

PR #383, opened 2024-09-23
Rename precheck module

mgoerens:
This PR proposes to rename the `precheck` module to `submission`.

This pertains to the upcoming refactoring that I intend to have ready (hopefully) very soon: As you know in the past few months, I created the `precheck` module and contributed code to it. This code is currently not being run in the pipeline, as I still needed some time to straighten things up and ensure end to end tests are passing. I'm getting closer.

The context behind this refactoring, as it might help better understand my will for a rename (besides not liking "precheck" even though I was the one introducing it):

The main idea is to better organize the CI pipeline, makes things more readable and less redundant, by introducing a job that runs before chart-verifier and that is responsible of performing sanity checks on the PR. Things like checking that the required files have been provided, chart version follows semver, web_catalog_only status, and so on. Currently some of those checks are performed in `checkprcontent` module, but also at different places in the codebase, and sometimes even twice. I'm in the process of moving all of those checks to one place. I'll provide more details in the PR that introduces the switch.

The (for now called) `precheck` modules introduces the data structure `Submission`, that represents a user's PR and associated information that pertains to either the certification of a chart, or the addition/modification of an OWNER file. The main purpose of this module is to confirm that the submission is "valid".

Hence my suggestion to have:
- the python module called `submission`.
- the upcoming file that will contain the `main` function called `validate.py`.
- the upcoming job that will run before `chart-verifier` in the CI pipeline called `validate-submission`.
- the binary (as defined in `setup.py`) called `validate-submission`.

Hopefully that makes sense. Even though we shouldn't take too much time arguing about naming (naming *is* hard), I also believe that names are important, so ... I'm open to other suggestions as well !

**EDIT:** For context, here is how the `build.yaml` workflow is likely to look like eventually: https://github.com/mgoerens/development/blob/switch_to_submission_validate/.github/workflows/build.yml


komish (approved):
/lgtm

PR #384, opened 2024-09-24
Assume webCatalogOnly=False if not found in OWNERS

mgoerens:
This mimics the current behavior, see: https://github.com/openshift-helm-charts/development/blob/3ac6de443b51d842c4438d9d4922b58b37bbe4c3/scripts/src/checkprcontent/checkpr.py#L36

When `get_web_catalog_only` is called without `raise_if_missing=True` it defaults to `False`.


komish (approved):
/lgtm

PR #385, opened 2024-09-24
Set version as optional when for Charts

mgoerens:
In the case where the submission only contains an OWNERS file, it does not relate to a specific chart's version. In particular, there is no version that can be inferred from the directory structure. Hence this should be an optional parameter when registering chart's information.


komish (approved):
/lgtm

PR #388, opened 2024-09-24
Match error message with end to end tests

mgoerens:
End to end tests expect specific messages to be present in the failed PR comments.

This PR aligns the error message from the submission module so that end to end tests may pass.

For instance: https://github.com/openshift-helm-charts/development/blob/main/tests/functional/behave_features/HC-06_provider_delivery_control.feature#L46-L47

### DIFF
--- a/scripts/src/submission/submission.py
+++ b/scripts/src/submission/submission.py
@@ -380,7 +380,7 @@ class Submission:
 
         return False, msg
 
-    def parse_web_catalog_only(self, repo_path=""):
+    def parse_web_catalog_only(self, repo_path: str = ""):
         """Set the web_catalog_only attribute
 
         This is achieved by:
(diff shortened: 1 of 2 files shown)


komish (commented):
Overall, looks fine. One thing to keep in mind, though is that we don’t technically need the exception to contain the E2E-relevant message. If we can catch the exceptions, we can emit the appropriate output to either E2E or whatever GitHub comment we want.

That in mind, would you still change this exception’s message? If so, I’ll throw the LGTM.

mgoerens:
Totally we could, that would mean having custom Exception for each possible error, so that we know which error message to emit. Currently I'm catching the exception, and using the Exception message as error message directly.

I don't know what's more pythonic / readable / understandable ? We can also discuss this in the upcoming refactoring PR to have the better picture. Right know, unless you have a strong opinion against my current approach I'd rather keep it.

komish (approved):
/lgtm

PR #389, opened 2024-09-24
Generate a valid empty index when none is found

mgoerens:
Also fixes the default github pages branch value.

Note that this was previously discussed [here](https://github.com/openshift-helm-charts/development/pull/347#discussion_r1630297722) and it was decided *against* initializing a empty index in that case.

However, the sandbox repo does not contain any index, hence, the end to end tests are failing when running this code.

Potentially this can be fixed by creating an empty (or not empty) index on the gh-pages branch of the sandbox repo, however I haven't looked into the other potential implication of doing so.

### DIFF
--- a/scripts/src/submission/submission.py
+++ b/scripts/src/submission/submission.py
@@ -518,15 +518,37 @@ def get_file_type(file_path):
     return "unknwown", None
 
 
-def download_index_data(repository, branch="gh_pages"):
-    """Download the helm repository index"""
-    r = requests.get(
-        f"https://raw.githubusercontent.com/{repository}/{branch}/index.yaml"
-    )
+def download_index_data(
+    repository: str, branch: str = "gh-pages", ignore_missing: bool = False
+) -> dict:
+    """Download the helm repository index
 
-    if r.status_code == 200:
-        data = yaml.load(r.text, Loader=Loader)
+    Args:
+        repository (str): Name of the GitHub repository to download the index file from.
+                            (e.g. "openshift-helm-charts/charts")
+        branch (str): GitHub branch to download the index file from. Defaults to "gh-pages".
+        ignore_missing (bool): Set to True to return a valid empty index, in the case the
+                            download of index.yaml fails.
+
+    Returns:
+        dict: The helm repository index
+
+    Raise:
+        HelmIndexError if the download fails. If not on the production repository, doesn't raise
+            and returns an empty index file instead.
+        HelmIndexError if the index file is not valid YAML.
+
+    """
+    index_url = f"https://raw.githubusercontent.com/{repository}/{branch}/index.yaml"
+    r = requests.get(index_url)
+
+    data = {"apiVersion": "v1", "entries": {}}
+    if r.status_code != 200 and not ignore_missing:
+        raise HelmIndexError(f"Error retrieving index file at {index_url}")
     else:
-        data = {}
+        try:
+            data = yaml.load(r.text, Loader=Loader)
+        except yaml.YAMLError as e:
+            raise HelmIndexError(f"Error parsing index file at {index_url}") from e
 
     return data


scripts/src/submission/submission.py
```diff
@@ -527,6 +527,6 @@ def download_index_data(repository, branch="gh_pages"):
     if r.status_code == 200:
         data = yaml.load(r.text, Loader=Loader)
     else:
-        data = {}
+        data = {"apiVersion": "v1", "entries": {}}
```
komish, line 531: One final thought here (not to rehash the previous discussion), do we want to add protections that would raise an exception in the production repo vs. non production repositories?

scripts/src/submission/submission.py
```diff
@@ -518,7 +518,7 @@ def get_file_type(file_path):
     return "unknwown", None
 
 
-def download_index_data(repository, branch="gh_pages"):
+def download_index_data(repository: str, branch: str = "gh-pages") -> dict:
```
komish, line 521: I assume this fix is specific to your submission module, and does not exist in the source that this reimplements? As in, there's no bug in the original code?

komish (changes requested):
Some small things noted inline.

scripts/src/submission/submission.py
```diff
@@ -527,6 +527,6 @@ def download_index_data(repository, branch="gh_pages"):
     if r.status_code == 200:
         data = yaml.load(r.text, Loader=Loader)
```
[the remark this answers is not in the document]
mgoerens: Done, check new code

scripts/src/submission/submission.py
mgoerens: I was hesitant. Because according to me the better solution would be to have some index file on the gh-pages branch of the sandbox repo so that end to end can pass. But I confess haven't taken the time to test the possible consequences of doing so (and am unsure which index file to craft exactly, though probably an empty index file set to `apiVersion": "v1", "entries": {}` would suffice.)

In the end I implemented the protection measure as it was easier to achieve.

scripts/src/submission/submission.py
mgoerens: The original code in checkpr doesn't use any default value. So this only for the submission module

scripts/src/submission/submission.py
komish: Hrm, two things now:

1. I lean towards not having a repository hard-coded in scripts here. I think what I'd rather see is a caller receive that context as an input, and _that input_ trigger the protection. We evaluate `$GITHUB_REPOSITORY` is not the production in various places in Actions, but we don't really see a lot of references to the repository information hard-coded to impact logic (validated with `grep`, so please cross-check me on this). Let me know what you think. I'm the fence, but I lean against having it hard-coded here, and having just something we toggle in the function signature that change sits behavior, and is determined by inputs when the script is called (e.g. CLI flags, environment variables, etc. that wrap this logic).

4. You probably don't want to raise an exception when the YAML loading fails, right? You just want to catch it, and then move on so you can return the empty index.  Something like this:

```python
...
try:
    data = yaml.load(r.text, Loader=Loader)
except yaml.YAMLError as e:
    # log e if you want to
    data = {"apiVersion": "v1", "entries": {}}
return data
```

komish: Whoops, I should have added this to a new review. These points apply to the changes as a whole. I've marked all other comments as resolved to clarify.

mgoerens: No problem, this way I can reply in the same thread and discussion isn't cluttered by linked PRs.

I pushed a change reg. (1), is this what you mean ? I would have to basically move the check to the GitHub action that runs the upcoming `validate.py` and add a CLI Flag (if ` $GITHUB_REPOSITORY="openshift-helm-charts/charts"`, then set a flag to False).

Reg. (2), in the event the function manages to fetch an index.yaml file from the repo/branch, but that index file is not valid ... Do we really want to default to an empty index ? I'd rather fail here. It means probably that something is wrong in our repo (Helm repository index is corrupted) or that the index got corrupted during download.

komish: Oh, duh! Thanks :laughing: Resolving this.

scripts/src/submission/submission.py
```diff
@@ -518,15 +518,37 @@ def get_file_type(file_path):
...
+        HelmIndexError if the index file is not valid YAML.
+
+    """
+    index_url = f"https://raw.githubusercontent.com/{repository}/{branch}/index.yaml"
+    r = requests.get(index_url)
 
-    if r.status_code == 200:
-        data = yaml.load(r.text, Loader=Loader)
+    data = {"apiVersion": "v1", "entries": {}}
+    if r.status_code != 200 and ignore_missing:
```
komish, line 546:
```suggestion
    if r.status_code != 200 and not ignore_missing:
```

komish (changes requested):
Last thing, I promise!

scripts/src/submission/submission.py
komish: I think this is what you meant?

mgoerens: Right ! Thanks

PR #390, opened 2024-09-24
Allow ignoring the presence of an OWNERS file

mgoerens:
When validating that a PR contains the correct set of file to certify a chart, this PR adds the possibility to ignore the fact that an OWNERS file is also present.

While an OWNERS file should never be present in this type of submission, the case of an invalid Submission containing a mix of OWNERS and other files is handled in the is_valid_owners_submission method.

Eventually, errors related to OWNERS file (which this belongs to) feed the owners-error-message GITHUB OUTPUT, while other errors feed the pr-content-error-message GITHUB OUTPUT.

EDIT: Current behavior can be checked in the checkpr module starting here: https://github.com/openshift-helm-charts/development/blob/3ac6de443b51d842c4438d9d4922b58b37bbe4c3/scripts/src/checkprcontent/checkpr.py#L120

### DIFF
--- a/scripts/src/submission/submission.py
+++ b/scripts/src/submission/submission.py
@@ -332,13 +332,20 @@ class Submission:
         else:
             self.modified_unknown.append(file_path)
 
-    def is_valid_certification_submission(self):
+    def is_valid_certification_submission(
+        self, ignore_owners: bool = False
+    ) -> tuple[bool, str]:
         """Check wether the files in this Submission are valid to attempt to certify a Chart
 
         We expect the user to provide either:
         * Only a report file
         * Only a chart - either as source or tarball
         * Both the report and the chart
+
+        Note: While an OWNERS file should never be present in this type of submission, the case of
+        an invalid Submission containing a mix of OWNERS and other files is handled in the
+        is_valid_owners_submission method. The flag "ignore_owners" allows for skipping this
+        speficic check.
 
         Returns False if:
         * The user attempts to create the OWNERS file for its project.
@@ -347,7 +354,7 @@ class Submission:
         Returns True in all other cases
 
         """
-        if self.modified_owners:
+        if self.modified_owners and not ignore_owners:
             return False, "[ERROR] Send OWNERS file by itself in a separate PR."
 
         if self.modified_unknown:
--- a/scripts/src/submission/submission_test.py
+++ b/scripts/src/submission/submission_test.py
@@ -281,6 +281,7 @@ class CertificationScenario:
     input_submission: submission.Submission
     expected_is_valid_certification: bool
     expected_reason: str = ""
+    ignore_owners: bool = False
 
 
 scenarios_certification_submission = [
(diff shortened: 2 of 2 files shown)


scripts/src/submission/submission.py
```diff
@@ -332,22 +332,29 @@ def set_tarball(self, file_path, tarball_match):
         else:
             self.modified_unknown.append(file_path)
 
-    def is_valid_certification_submission(self):
+    def is_valid_certification_submission(
+        self, ignore_owners: bool = False
+    ) -> tuple[bool, str]:
         """Check wether the files in this Submission are valid to attempt to certify a Chart
```
komish, line 338:
```suggestion
        """Check whether the files in this Submission are valid to attempt to certify a Chart
```

scripts/src/submission/submission.py
```diff
@@ -332,22 +332,29 @@ def set_tarball(self, file_path, tarball_match):
...
 
         We expect the user to provide either:
         * Only a report file
         * Only a chart - either as source or tarball
         * Both the report and the chart
 
+        Note: While an OWNERS file should never be present in this type of submission, the case of
+        an invalid Submission containing a mix of OWNERS and other files is handled in the
+        is_valid_owners_submission method. The flag "ignore_owners" allows for skipping this
+        speficic check.
```
komish, line 348:
```suggestion
        Note: While an OWNERS file should never be present in this type of submission, the case of
        an invalid Submission containing a mix of OWNERS and other files is handled in the
        is_valid_owners_submission method. The flag "ignore_owners" allows for skipping this
        specific check.
```

komish (commented):
I'm understanding that _another check_  is asserting if the submission is a valid operation on an OWNERS file (e.g. has only the OWNERS file and no other operations).

But, ensuring an OWNERS file is _not_ present is critical for a certification submission - in what case would we both run this code, AND not care about the presence of an owners file?

mgoerens:
(Note that I will refrain from pushing the typo correction for now, as to avoid having this discussion buried under 

> I'm understanding that another check is asserting if the submission is a valid operation on an OWNERS file (e.g. has only the OWNERS file and no other operations).

Yes another check exists here: https://github.com/openshift-helm-charts/development/blob/4e5253bb0e1377dd1b09cafd40f5bf06875f6d35/scripts/src/submission/submission.py#L372

> But, ensuring an OWNERS file is not present is critical for a certification submission - in what case would we both run this code, AND not care about the presence of an owners file?

This is a very valid question, but it's difficult to answer without using code that isn't already present.

In short, the upcoming `submission_validate` main function will have to create two github outputs (in order to mimic the exising `checkpr` behavior):
- `owners-error-message` -> This is populated as soon as an OWNERS file is present. This is using `is_valid_owners()`.
- `pr-content-error-message` -> Populated if something (1) is wrong with the PR content. **Crucially, it doesn't care about the presence of an OWNERS file in the list of modified files.** I've used, among other things, `is_valid_chart_certification()` to craft this message as it contains several relevant tests. But as errors related to OWNERS file don't end up in `pr-content-error-message`, I need to skip one check.

Note that the concepts of "OWNERS PR" and "Certification PR" aren't formally present in our codebase so far AFAIK. The addition of the `is_valid_owners_submission()` and `is_valid_certification_submission()` (when I started working on the `submission` module while not actually using it) were first only added to have a more formal definition of what files should be submitted in each cases. I admit it wasn't clear to me how exactly I would end up using them, but felt and still feel that it makes sense to have these concepts clearly defined as code.

Now I realize that there is actually no need in the upcoming `submission-validate` job to run `is_valid_certification_submission()` with `ignore_owners=False`, so I could simply remove this check. However, that would break the semantic for this function, as it would not actually provide the ability to check whether the file added by the PR fit the definition of a "valid Chart certification PR".

Hope this explanation makes sense.

(1) Full list of reasons that can populate the `pr-content-error-message`:
- the required files are not present (at least a report, a chart's source, or a tarball)
- web_catalog_only is inconsistent / incorrect between report and OWNERS
- the chart is already present in the Helm repository index or the release tag already exists
- the PR contains multiple charts, the tarball is incorrect, or the version is not semver compatible

komish:
Understood. It's reasonable to reconsider this piece as a part of the whole picture. That in mind, feel free to submit typo fixes in the next PR. We don't need to retrigger this pipeline just for that right now.

/lgtm

komish (approved):

PR #392, opened 2024-10-01
Extract PR validation from chart-verifier job

### LINKED ISSUE #296: Create extra job for extracting and validating all needed info from PR

mgoerens:
This PR refactors the certification pipeline by splitting the `chart-verifier` job into three jobs, running sequentially:
- The new `validate-submission` is responsible for extracting and validating all information related to a PR.
- The existing `chart-verifier` job is now only responsible for running chart-verifier against the Chart source (if provided, or verifies the submitted report.yaml instead).
- The new `manage-gh-pr` adds a comment on the PR, and merges it in case the pipeline was successfull.

This PR also introduces a new mechanism for passing information between jobs. In addition to the existing GitHub outputs, a submission.json artifact containing a json representation of the Submission object is created in `validate-submission`. Subsequent steps/jobs now have the possibility to read from this file.

The main pursued goal behind this PR is to better separate concerns and avoid mixed up logic, thus improving readability and clarity of our pipeline and improve future maintainability and onboarding.

This PR also opens up the ability to reducde code deduplication and redundant checks. In the interest of keeping this PR to a minimum, this has not been fully done here. One example: `get-verify-params.py` currently queries the GitHub API in order to get chart's information, while they are already available in the submission.json artifact. This rather trivial change is left for a future PR.

Note that the `submission` python module is essentially replacing the old `checkpr` module. The only function remaining used by other modules being `get_file_match_compiled_patterns`, it is moved to the 'reporegex' module.

Closes #296


jsm84:
@mgoerens other than that one issue with the sandbox repo URL, everything else looks good.

All tests completed against my own infrastructure:
* forked dev repo (based on this PR branch)
* separate 3-node cluster
* forked sandbox
* separate bots/tokens for both forked repos

I'll approve once the above issue is fixed (admittedly minor, and only affects forked sandbox repos).

jsm84 (changes requested):
@mgoerens Sorry for not posting this...

scripts/src/submission/submission.py
```diff
@@ -237,7 +236,6 @@ def __post_init__(self):
         if not self.modified_files:
             self.modified_files = []
             self._get_modified_files()
-            self._parse_modified_files()
```
komish, line 240: We no longer need to parse files here?

Also, in the next couple of lines (I can't select it here for some reason) - is there a reason why `files_api_url` is set using re.sub now?

scripts/src/submission/submission.py
```diff
@@ -535,8 +533,11 @@ def is_valid_web_catalog_only(self, repo_path: str = "") -> tuple[bool, str]:
 
         return True, ""
 
+    def get_pr_number(self):
+        return self.api_url.split("/")[-1]
 
-def get_file_type(file_path):
+
+def get_file_type(file_path: str) -> tuple[str, re.Match[str]]:
```
komish, line 540: - The return value of this function has a typo ("Unknwown"). I can't select it for this comment because you didn't modify it here.

komish (changes requested):
Overall, I like the idea of separating the concerns of parsing pull request content. Comments left in line for things that I believe are worth considering. Generally, the Submission lib replacing the previous checkprcontent gets a +1 from me.

One design challenge I have is with the flow of the build.yml task. A PR's content is effectively its gate. In my mind, there's effectively no reason to run chart-verifier (or make any concessions such as redhat-to-community, etc.) if the content is considered invalid for any reason. 

Previously, [that was captured clearly in this step](https://github.com/openshift-helm-charts/development/blob/main/.github/workflows/build.yml#L179-L183), where the PR content's invalidity explicitly exits the pipeline.

These changes appear to be implement the same. [This](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L198) conditional prevents chart-verifier from running if the validation step is not successful, as expected). And [this](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L404) condition will run the "PR commenting logic" in all cases, even if the update would contain no verifier content, and only the PR validation content.

What I feel like I'm lacking is some clarity, and I think [this](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L446-L459) is a big part of it. This looks to me like it was a carryover from the previous build.yml, but because validation and and chart-verification execution are now _separate_ jobs, this manage-gh-pr task feels like it's doing too much. It's handling the commenting in all cases (whether validation passed/failed, and/or whether chart-verifier passed/failed) which feels like it blurs the line between the PR content validation task being a hard-stop to the workflow. This could potentially exacerbate the other comments I made inline re: conditional logic going away in favor of implicit (to GitHub Actions) logic handling those same conditions.

Could just be me, so I'd be curious to get @jsm84's take on it as well as yours.

.github/workflows/build.yml
```diff
@@ -137,21 +143,31 @@ jobs:
...
+      - name: Extract PR information
+        id: validate-submission
         env:
-          GITHUB_REF: ${{ github.ref }}
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
         run: |
-          INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "refs/heads/gh-pages"; else echo "${GITHUB_REF}-gh-pages"; fi)
-          ./ve1/bin/check-pr-content --index-branch=${INDEX_BRANCH} --repository=${{ github.repository }} --api-url=${{ github.event.pull_request._links.self.href }}
+          # Sandbox repo does not have an helm repository index
+          if [ "$GITHUB_REPOSITORY" == "openshift-helm-charts/sandbox" ]; then
```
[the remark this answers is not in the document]
jsm84: np, that works!

mgoerens:
I've addressed most comments in a fixup commit to make review easier.

To answer @komish comment:

> Overall, I like the idea of separating the concerns of parsing pull request content. Comments left in line for things that I believe are worth considering. Generally, the Submission lib replacing the previous checkprcontent gets a +1 from me.
> 
> One design challenge I have is with the flow of the build.yml task. A PR's content is effectively its gate. In my mind, there's effectively no reason to run chart-verifier (or make any concessions such as redhat-to-community, etc.) if the content is considered invalid for any reason.
> 
> Previously, [that was captured clearly in this step](https://github.com/openshift-helm-charts/development/blob/main/.github/workflows/build.yml#L179-L183), where the PR content's invalidity explicitly exits the pipeline.
> 

While explicit, there are still a bunch of steps (with the `always` condition) that run after this one even in the event of a failure. Especially, the ones that are related to commenting the PR, communicating the error message to the end user.

See [this PR](https://github.com/openshift-helm-charts/sandbox/actions/runs/11456373433/job/31874387401?pr=106086) for example.

> These changes appear to be implement the same. [This](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L198) conditional prevents chart-verifier from running if the validation step is not successful, as expected). And [this](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L404) condition will run the "PR commenting logic" in all cases, even if the update would contain no verifier content, and only the PR validation content.
> 

Yes, and that is a clearer implementation IMO. FYI, [this is how it looks now](https://github.com/openshift-helm-charts/sandbox/actions/runs/11437593388/job/31817434385?pr=106016). It's clear that the chart-verifier part is totally skipped, also it avoids having to play with `continue-on-error` and `always` conditions too much, which to me feels difficult to read in the current situation.

But I had my head down into this work for some time so it's likely I'm biased towards the implementation I went with. I appreciate your critical external point of view !

> What I feel like I'm lacking is some clarity, and I think [this](https://github.com/openshift-helm-charts/development/blob/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba/.github/workflows/build.yml#L446-L459) is a big part of it. This looks to me like it was a carryover from the previous build.yml, but because validation and and chart-verification execution are now separate jobs, this manage-gh-pr task feels like it's doing too much. It's handling the commenting in all cases (whether validation passed/failed, and/or whether chart-verifier passed/failed) which feels like it blurs the line between the PR content validation task being a hard-stop to the workflow. This could potentially exacerbate the other comments I made inline re: conditional logic going away in favor of implicit (to GitHub Actions) logic handling those same conditions.

Here I think I need a discussion because I don't see how this could be implemented differently. Sure I could add a step that explicitly exits after `validate-submission` fails, but still we'd need to add a comment to the PR.

That is "handles the commenting in all cases" doesn't really bothers me.

We could look into separating commenting (always) and merging (only if all previous jobs succeeded) the PR into two different jobs for more clarity.

> 
> Could just be me, so I'd be curious to get @jsm84's take on it as well as yours.

komish:
> We could look into separating commenting (always) and merging (only if all previous jobs succeeded) the PR into two different jobs for more clarity.

This could help. But as I look into this more, I'm not convinced it's truly needed. Might be okay with leaving it as is. Let's sync up some time mid next week to discuss. I'm working through the library's semantics hands on, and I should have some more to offer by then.

komish (approved):
/lgtm

jsm84 (approved):
/ltgm

Issue #327, opened 2024-02-14
Explore using a matrix in GitHub actions to call individual tags during E2E testing


komish:
One of the biggest challenges we run into with E2E testing is with transient errors (e.g. GitHub rate limits). One test case failing requires us to re-run the entire E2E workflow because we currently call *all* BDD tests in a single call.

Explore the use of GitHub's [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) functionality to call the full test suite as individual jobs. Explore concurrency, but keep in mind that we don't want the cluster to tumble because we're running multiple E2E test suites at once.

komish:
I reconfigured testing for #359 to shorten the feedback loop, and it allowed some basic testing of matrix based execution. Here's the workflow file that I used.

```yaml
name: "Matrix-based Testing"

on:
  workflow_dispatch:

jobs:
  get-features:
    runs-on: ubuntu-latest
    outputs:
      features: ${{ steps.find-features.outputs.features }}
    steps:
      - uses: actions/checkout@v4
      - name: find features
        id: find-features
        run: |
          cd tests/functional/behave_features
          echo features=$(find . -name '*.feature' | sed -e 's%\./%%g' | jq -R -s -c 'split("\n")[:-1]') | tee -a $GITHUB_OUTPUT
  run-tests:
    runs-on: ubuntu-latest
    needs: [get-features]
    strategy:
      fail-fast: false
      max-parallel: 4
      matrix: 
        feature-file: ${{ fromJson(needs.get-features.outputs.features) }}
    steps:
      - name: Checkout Base Branch
        uses: actions/checkout@v4
        with:
          token: ${{ secrets.BOT_TOKEN }}

      - name: Set up Python 3.x Part 1
        uses: actions/setup-python@v5
        with:
          python-version: "3.10"

      - name: Set up Python scripts
        run: |
          # set up python scripts
          echo "set up python script in $PWD"
          python3 -m venv ve1
          cd scripts
          ../ve1/bin/pip3 install -r requirements.txt
          ../ve1/bin/pip3 install .
          cd ..
      - name: Test CI Workflow
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          BOT_NAME: ${{ secrets.BOT_NAME }}
          BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
[...]

PR #397, opened 2024-11-06
Revert nightly blockers

komish:
This refactor is being temporarily reverted so that we can roll out bugfixes and environmental changes. 

https://github.com/openshift-helm-charts/development/issues/327 (or similar) will be prioritized, after which these changes will be reintroduced.


jsm84 (approved):
/lgtm

PR #398, opened 2024-11-06
fix missing RELEASE_TAG value for index update commits

komish:
Index update commit messages no longer have the chart's release information (i.e. org-chartname-version).  This is because there is data not being transferred across jobs in the certification workflow.

The `RELEASE_TAG` environment variable is used in this script, but has no value assigned to it. It's being set in a previous job, so we can easily pass along this variable to this script.


mgoerens (approved):
/lgtm

mgoerens:
How did this come to not working ? What changed ?

komish:
@mgoerens the value was previously calculated in-script, vs. being passed in as an environment variable. When the shell script was introduced (replacing the python script that preceded it), the environment variable wasn't added to the task, but empty vars by default are not an error in Bash, so it was quietly "failing".

PR #401, opened 2024-11-19
Cache virtualenv between jobs

mgoerens:
Installation of the python dependencies and scripts is done at the start of each job within the build.yaml workflow.

By caching the venv we can save up to 10 seconds on each job after the first.

**NOTE FOR REVIEWERS:** This builds upon #419 and should hence be merged after. Currently it also contains the commit of #419 but we should rebase before merging.

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -5,9 +5,14 @@ description: |
 
   Serves as the single place to update python versions over time across the
   project.
+outputs:
+  python-version:
+    description: Version of Python installed
+    value: ${{ steps.setup-python.outputs.python-version }}
 runs:
   using: composite
   steps:
   - uses: actions/setup-python@v5
+    id: setup-python
     with:
       python-version: '3.10'
--- a/.github/actions/install-ci-scripts/action.yml
+++ b/.github/actions/install-ci-scripts/action.yml
@@ -1,0 +1,33 @@
+name: Install our CI scripts
+description: |
+  Handles installation and caching/restore of the CI scripts in a virtualenv.
+inputs:
+  working-directory:
+    description: Basedir in which to save / restore cache and install venv.
+    default: './'
+    required: false
+  python-version:
+    description: Installed Python version
+    required: true
+runs:
+  using: composite
+  steps:
+  # This attempts to restore the virtualenv from a GitHub cache with the corresponding cache key.
+  # On cache-miss, this automatically adds a post step that creates a cache from this path and using this cache key.
+  - name: Cache virtualenv
+    id: cache
+    uses: actions/cache@v4
+    with:
+      key: venv-${{ runner.os }}-${{ inputs.python-version }}-${{ hashFiles('scripts/requirements.txt') }}
+      path: ${{ inputs.working-directory }}/ve1
+
+  - name: Setup venv on cache-miss
+    if: steps.cache.outputs.cache-hit != 'true'
+    shell: bash
+    working-directory: ${{ inputs.working-directory }}
+    run: |
+      # set up python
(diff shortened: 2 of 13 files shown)


komish (commented):
I'm not sure we need to call explicit save/restore cache actions for what you're looking to achieve here.

IIRC, calling actions/cache will automatically cache the content at the specified paths, assume you've generated it, to the specified cache key.

https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action

What this should let you do is go through the following process every time you need to restore the cache:

- try and restore
- on cache-miss, build
- ... other tasks using the content...
- if cache-miss, save built content to cache

The issue I see with the current implementation is that we could end up in a deadlock if for some reason the cached content disappears between the step that saved the content and the step(s) that tried to restore the content. 

If instead we were to try to restore, and on miss, build, the workflow should be able to complete regardless of the state of the cache.

Let me know what you think.

mgoerens:
@komish Please have another look

I've implemented something close to what you were suggesting:
- Try and restore cached venv
- If cache-miss, build venv, and save cache
- [... rest of the workflow ... ]

I believe it makes sense to add this logic to the `setup-python` action. Let me know if you prefer to keep this action as simple as possible. As a nice side effect, this allows this cache mechanism to be applied to every workflow that uses this action, rather than limiting it to the build.yaml workflow.

.github/actions/setup-python/action.yml
```diff
@@ -5,9 +5,45 @@ description: |
...
+    required: false
 runs:
   using: composite
   steps:
   - uses: actions/setup-python@v5
+    id: setup-python
     with:
       python-version: '3.10'
+
+  - name: Attempt to restore cached virtualenv
```
komish, line 25: I feel script installation would be better off in its own composite actions, and have the caching performed there. I know that in most of our use cases, we'll install python && install the CI scripts in their virtualenv, but I think ultimately that these things should live separately and have their inputs populated as needed.

If you want to then create a "common path" that reuses the two, that's fine.

Making some assumptions, I would guess that part of the reason why these two are together in this implementation is because `${{ steps.setup-python.outputs.python-version }}` is being used as part of the cache-key. I'd say if you *must* have this value and can't assume one accept this as a required input to the reusable action.

Also, I do see you using explicit cache/restore and cache/save actions - and I'd assume that's because there are additional conditions that would control whether or not caching is to be performed. Just want to confirm that this is the reason driving that design (vs. using the cache action itself and letting it handle save/restore).

---

To summarize, do you think it's worth designing an "install-ci-scripts" (this name is just an example) composite action that includes the caching logic you want, handled nicely during the installation of the venv?

If so, one case that we want to make sure we consider here is when we install venvs from pr-branch vs. the default clone in a pipeline.

.github/actions/setup-python/action.yml
```diff
@@ -5,9 +5,45 @@ description: |
 
   Serves as the single place to update python versions over time across the
   project.
+inputs:
+  skip-venv-install:
```
komish, line 9: NIT: it might not matter based on other comments in this review, but I'd rather see this as `install-venv` rather than `skip-install`.

komish (changes requested):
A couple of more comments, now, as the design has changed a little bit.

Also! With long PRs like these, we've discussed how hard it is to see code review comments because of the E2E testing being extra noisy and polluting the UI. When you look at the Files Changed tab, though, there's a little "Conversations" drop down that lets you see in-code comments, which might be easier to navigate.

It's possible we talked about this already,  but mentioning again as navigating this UI has been challenging as always :laughing: 

Thanks for your work on this @mgoerens.

PR #402, opened 2024-11-19
Install oc without external action

mgoerens:
The redhat-actions/openshift-tools-installer is experiencing slowness when installing oc. This steps can take up to 3 minutes. More info: https://github.com/redhat-actions/openshift-tools-installer/issues/105

Using a custom script reduces the installation of oc to less than 10s.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -267,9 +267,11 @@ jobs:
       - name: Install oc
         id: install-oc
         if: ${{ steps.verify_requires.outputs.cluster_needed == 'true' }}
-        uses: redhat-actions/openshift-tools-installer@v1
-        with:
-          oc: latest
+        run: |
+          wget --quiet https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz
+          tar -zxvf openshift-client-linux.tar.gz oc -C /usr/local/bin/
+          rm -f openshift-client-linux.tar.gz
+          oc version --client
 
       - name: Set cluster login params
         id: login-params


.github/workflows/build.yml
```diff
@@ -216,9 +216,12 @@ jobs:
       - name: Install oc
         id: install-oc
         if: ${{ steps.verify_requires.outputs.cluster_needed == 'true' }}
-        uses: redhat-actions/openshift-tools-installer@v1
-        with:
-          oc: latest
+        run: |
+          wget --progress=dot:mega https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz
```
komish, line 220: Probably don't need `--progress` in CI.

.github/workflows/build.yml
```diff
@@ -216,9 +216,12 @@ jobs:
...
+          tar -zxvf openshift-client-linux.tar.gz oc -C .
+          echo "$PWD/oc" >> $GITHUB_PATH
```
komish, line 222: You prefer this to symlinking to /usr/local/bin/?

.github/workflows/build.yml
```diff
@@ -216,9 +216,12 @@ jobs:
...
+          rm -f openshift-client-linux.tar.gz
+          oc version
```
komish, line 224: Pass the --client flag here. We don't care about server details here.

komish (changes requested):
1). Remove progress bar flags

2). Should we consider making this a reusable action? I know we also pull `oc` down in chart-verifier's CI. https://github.com/redhat-certification/chart-verifier/blob/main/.github/workflows/build.yaml#L83-L91

3). Does adding this to $GITHUB_PATH add it to the end or the beginning of the resolution? Just want to make sure we won't get an `oc` that comes up sooner in the path than what we're installing.

mgoerens:
@komish First off, I want to briefly address the pertinence of this PR. Do we want to (a) implement this workaround, or (b) investigate the upstream action ?

In favor of (a):
- it's relatively simple to do (in fact this is as good as done, even though your other comments are valid) 
- the upstream issue is not trivial, also the upstream action is not actively maintained so we would are almost on our own.
- (previously this issue that this is solving was critic as we had very frequent CI timeouts) -> less relevant today.

In favor of (b):
- fixing it upstream means fixing it for everybody
- @komish has some experience with the issue and with the upstream repo as he already fixed a similar issue.

---

I will answer your other comments at a later time

.github/workflows/build.yml
mgoerens: Edit: Found the `--quiet` option !

---

Here is how it looks like if you don't specify a progress option: https://github.com/mgoerens/test-github-workflow/actions/runs/13547227001/job/37861759632

I've settled for `--progress=dot:mega`: https://github.com/mgoerens/test-github-workflow/actions/runs/13547210704/job/37861708301

But we can also go to `--progress=dot:giga`: https://github.com/mgoerens/test-github-workflow/actions/runs/13547548815/job/37862815910

There's no way to completely remove progress outputs AFAIK. Note that Github actions (or at least the `ubuntu-22.04` image this runs on) are using wget 1.21.2, while my Fedora has wget 2.2.0 which seems to have a different set of options.

komish (approved):
/lgtm

PR #403, opened 2024-11-25
implement matrix-testing with a reusable behave workflow

### LINKED ISSUE #327: Explore using a matrix in GitHub actions to call individual tags during E2E testing

komish:
Fixes #327 

**Summary - this PR:**

1. Adds a net-new reusable workflow (behave.yml) that will run our behave-driven tests, such as the nightly and smoke tests in a matrix, subdivided by feature file definition.
2. Removes version-check related tests and code that enabled those tests. As discussed with maintainers, this check did not seem to deliver as much value and added a good deal of overhead in behave testing that only applied to that use case.
3. Refactors all behave testing callers to use the new reusable workflow. In effect, this amounts to the nightly tests, release tests, and smoke tests used for workflow modification pull requests.

**Long form notes for maintainers:**

- Reusable workflow vs. composite actions: I ended up using a reusable workflow instead of a composite action because the former gives access to the user's base environment. I believe I also read that you cannot run a matrix in a composite action, and instead need to invoke the matrix at the call point. This would have added some overhead to every call point that just seemed inefficient. The reusable workflow completely encapsulates the behave tests, so the caller just needs to specify what tags and the state of the repository (if not default) to pull.

- Splitting single-job workflows into multiple jobs:  Arguably the trickiest part of this effort, I had to split workflows into, at the very least, two jobs (from one) because a matrix happens at the job level. In effect, the task that would run in a matrix would inevitably need to be in its own job, and therefore would have to receive both inputs from any previous jobs, and emit outputs for subsequent jobs. For this reason, there are several changes that partition out workflows that would be one or two jobs into many jobs.

- Checkout-related inputs to behave.yml: In certain cases, the behave.yml would need to pull the changes submitted by the contributor in order to either (a) use new scripts, or (b) test updated workflows. Most of the workflows that call behave.yml are pull_request_target-triggered workflows, and so the actions/checkout task does not pull the PR (for security reasons) but instead pulls the HEAD of the main branch. We gate pull requests for critical things like this using the check-contributor action in-repo, and release-related scripts and entrypoints check it as well. Either way, in order to run the new code, we have to be able to pull the new code in. The inputs in behave.yml have had descriptions added so that the caller understands their responsibility in checking the author of the content before running.

- Using a list of feature files as the matrices: IIRC I couldn't tell behave to run only HC-001 features without writing a tag for each new feature file. I also couldn't find a a clean way to generate a list of tests that were both in the full tag AND the HC-001 feature definition. Either way, both of those options would have required us to modify our tag definitions in workflow every time we want to either extend a feature file or add new features. To that end, it felt better to auto-generate the list of files and just run the full tag against the individual feature file. Worth noting - when you specify a feature file with a tag, and that feature file has no tests with that tag, you'll just get an auto-green test case in GitHub Actions. It's not perfect, but it's the simplest way I've been able to split the work into a matrix. Open to suggestions if there is an obvious other approach that doesn't have similar (or its own array of alternate) tradeoffs.

**Where can I see it running?**

Here's an example release: https://github.com/practice-room/development/actions/runs/12019147275/job/33508212448

Here's an example of the PR to the production repo containing new workflows and scripts. The test set here has been reduced to speed up the development feedback cycle. That's why we only see two tests: https://github.com/practice-room/charts/pull/9

**Remaining concerns**:

- This should speed up the behave full test suite a reasonable amount, but each new job that we spin up means we have to get scheduled to a runner which adds work time. Hopefully the overall time will be improved compared to our previous 5-6 hours. We can continue to explore other ways to reduce this time (such as reducing the timeout on our 30 minute timeout tests).
- We have to be careful with secondary rate limits now. The number of API requests made per minute may be high in release cases. E.g. after we've sent a bunch of requests for a release, then have PRs against all three repositories that may continue to count against API limits. For this reason, concurrency of the matrix is set to 4. As long as this value isn't 1, I'd think we should see an improvement overall.


.github/workflows/test.yml
```diff
@@ -84,6 +67,11 @@ jobs:
           # workflow only change but user not authorized
           exit 1
 
+      # BUG: This task attempts to run the `full` behave tag if the PR under
+      # test is a release from dev to prod, but the matcher condition that would
+      # emit this appears broken. Investigate the setting of the
+      # charts_release_branch output, or just run smoke tests and remove the
+      # condition associated with this output.
```
mgoerens, line 74: Should we create an issue in our repo to track this ?

I'm making some sense out of this comment, but capturing the full context right now would be beneficial.

.github/workflows/test.yml
```diff
@@ -104,60 +92,63 @@ jobs:
...
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          echo "Not a release PR from dev to charts, so running only smoke tests"
+          echo "test-tags=smoke" | tee -a $GITHUB_OUTPUT
+  
+  run-tests:
+    # No further pull request author checking done here because
+    # check-contributor gates the needed jobs.
+    name: Run Tests
+    needs:
+    - determine-workflow-conditions
```
mgoerens, line 115: I know it doesn't really change anything, but should it also need `check-contributor` to make the running order more explicit ?

I'm not sure what the best practice is (specifying all jobs that have to run before vs specifying only the previous job), but this is usually what I've seen in our repo.

.github/workflows/test.yml
```diff
@@ -104,60 +92,63 @@ jobs:
...
-        uses: pascalgn/automerge-action@v0.16.2
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          MERGE_METHOD: squash
-          MERGE_LABELS: ""
+  approve-and-merge:
+    name: Approve and merge
+    needs:
+    - determine-workflow-conditions
+    - run-tests
```
mgoerens, line 137: Same here wrt needs `check-contributor`.

Actually the same could be asked for the jobs in the release.yml workflow

.github/workflows/release.yml
```diff
@@ -150,38 +193,56 @@ jobs:
...
+    outputs:
+      should-merge: ${{ steps.check_merge_and_release.outputs.merge }}
+      release-tag: ${{ steps.check_merge_and_release.outputs.release_tag }}
+      release-body: ${{ steps.check_merge_and_release.outputs.release_body }}
+    runs-on: ubuntu-22.04
+    # cancelled() required because jobs this workflow depends on can skip.
+    if: |
+      !cancelled() &&
+      needs.determine-workflow-conditions.result == 'success' &&
+      ( needs.sync-repos.result == 'success' || needs.sync-repos.result == 'skipped' )
```
mgoerens, line 235: I'm having trouble understanding the need for `!cancelled()` and `( needs.sync-repos.result == 'success' || needs.sync-repos.result == 'skipped' )`)

> Possible values [of a job result] are success, failure, cancelled, or skipped.

A job that is skipped (i.e. the attached `if` condition evaluates to `false`) will have its result set to `skipped`. This will not prevent subsequent job to run. Here `( needs.sync-repos.result == 'success' || needs.sync-repos.result == 'skipped' )` feels redundant with the default behavior.

Similarly, the job will not run if the workflow was cancelled. In other words, it will run if the workflow was not cancelled. Again, `!cancelled()` feels redundant.

Maybe I'm missing something, what are we trying to achieve here ?

.github/workflows/release.yml
```diff
@@ -197,48 +258,80 @@ jobs:
...
-          github-token:  ${{ secrets.GITHUB_TOKEN }}
+  approve-merge-release:
+    name: Approve, merge, release
+    needs: [determine-merge-and-release-conditions]
+    runs-on: ubuntu-22.04
+    # cancelled() required because jobs this workflow depends on can skip.
+    if: |
+      !cancelled() &&
+      needs.determine-merge-and-release-conditions.outputs.should-merge == 'true' ||
+      needs.determine-merge-and-release-conditions.outputs.release-tag != ''
```
mgoerens, line 283: Same question wrt `!cancelled()`

mgoerens (commented):
Thanks for your work, this will greatly improve our tests.

No blockers from my side, but I have added some questions and possible improvements.

Regarding your notes:

> This should speed up the behave full test suite a reasonable amount, but each new job that we spin up means we have to get scheduled to a runner which adds work time. Hopefully the overall time will be improved compared to our previous 5-6 hours. We can continue to explore other ways to reduce this time (such as reducing the timeout on our 30 minute timeout tests).

Possible improvement would be to better manage the HC16 tests as you mention. They run 3 tests in sequence, that time out after 30 minutes, totaling to over 90 minutes for only one feature while all other seem to take max 20 minutes.

Either:
- Finding a way to call them in different jobs, so that they potentially can run in parallel. Either by breaking the feature into 3, or by finding a way to call each tests individually.
- As you mention, reducing the timeout to something less than 30 minutes. But that would stop reflecting "the real world".
- We could also try to see if we could prioritize this feature so that it takes a "thread" from the start.

In any case it's not a blocker for this work.

> We have to be careful with secondary rate limits now. The number of API requests made per minute may be high in release cases. E.g. after we've sent a bunch of requests for a release, then have PRs against all three repositories that may continue to count against API limits. For this reason, concurrency of the matrix is set to 4. As long as this value isn't 1, I'd think we should see an improvement overall.

Do we have a way to monitor our GH API usage ? Do we know how close we are from the limits ?

komish:
@mgoerens thanks for looking

I'll respond to in-line code directly, but on the top-level review comments, just a couple of notes:

> Possible improvement would be to better manage the HC16 tests as you mention. They run 3 tests in sequence, that time out after 30 minutes, totaling to over 90 minutes for only one feature while all other seem to take max 20 minutes.

^ I've created https://github.com/openshift-helm-charts/development/issues/407 to track this, and have copied your comments over. I do think I'm open to reducing the real-world parallel here, or otherwise reducing the chart-verifier default timeout (or even in-repo timeout) to something more reasonable. Reorganizing the features seems okay as well, but I'd still like it to be shorter. I can't imagine many E2E test cases are taking the full 30m to install the application and run chart-testing.

> Do we have a way to monitor our GH API usage ? Do we know how close we are from the limits ?

I don't think we monitor it directly, but we do get log lines emitting the API limits remaining in various places. Might be a good enhancement opportunity.

.github/workflows/test.yml
komish: Good catch. This is similar to the discussion we had [here](https://github.com/openshift-helm-charts/development/pull/392/files/6a88a5d2998adb24c637381da8cb7ab6a15cb3ba#r1806729666) and I'm open to all options here.

GitHub "puts these in a line" from a dependency perspective, because *determine-workflow-conditions* needs *check-contributor*. Beyond documentation clarity: E.g. "this job should be contributor-gated, and is as written", the only other reason to add something to the *needs* list from GitHub's perspect is to use its outputs as inputs to the dependent job.

From a documentation perspective, I've added two lines of comment for something that could have been one line of comment + an entry in the *needs* list. With that in mind, I'm open to doing this either way. It wouldn't change the execution as far as I can tell.

.github/workflows/test.yml
```diff
@@ -67,14 +56,8 @@ jobs:
...
-          echo "Request type: '$GITHUB_EVENT_NAME'"
-          if [ "$GITHUB_EVENT_NAME" == "pull_request_target" ]; then
-              echo "[INFO] check if PR contains only workflow changes and user is authorized"
-              ve1/bin/check-pr-for-ci --verify-user=${{ github.event.pull_request.user.login }} --api-url=${{ github.event.pull_request._links.self.href }}
-          else
-              echo "[INFO] manual invocation - check if user is authorized"
-              ve1/bin/check-pr-for-ci --verify-user=${{ github.actor }}
-          fi
+          echo "[INFO] check if PR contains only workflow changes and user is authorized"
+          ve1/bin/check-pr-for-ci --verify-user=${{ github.event.pull_request.user.login }} --api-url=${{ github.event.pull_request._links.self.href }}
```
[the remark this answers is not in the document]
komish: Test Workflow was pulling double duty. If you triggered a run via Workflow Dispatch, you would effectively be running the version-change behave tag, without the bits that would update the software-version branch of the repository, etc. As those operations are being removed the Workflow Dispatch element of this test needs to go as well.

> If we don't forecast calling this script without --api_url anymore, then we should probably set this argument to required=True and remove the logic that follows if not args.api_url.

Happy to do this, but I'm not 100% that we'll never use that logic again. As an aside, I do think that logic could use a refactor in general (given that the user verification piece seems repeated, but I digress. Let me know if you have a strong opinion one way or the other.

.github/workflows/release.yml
komish: This job needs **sync_repos**, which can skip. If it skips, forcing the execution of this job means that `needs.sync_repos.outputs` will be empty, which we want[1]. The calling code already expects the empty values.

> A job that is skipped (i.e. the attached if condition evaluates to false) will have its result set to skipped. This will not prevent subsequent job to run

In my testing, having a result set to skipped **will** prevent a subsequent job from running if it's defined as a dependency via needs. See [this action run](https://github.com/komish/gh-actions-hackery/actions/runs/12124329009) and its accompanying [workflow file](https://github.com/komish/gh-actions-hackery/blob/f10a9db2d91995cf944acec91308d8a68c4f51a2/.github/workflows/skip-behavior.yml). In this case, "three" needs "two" which skipped, so it doesn't run.

Further, even if the Job in question does NOT depend on a skippable job, if it depends on a descendant of a skippable job, and the skippable job skips, it will also not run. See [this action run](https://github.com/komish/gh-actions-hackery/actions/runs/12124141347)  and its accompanying [workflow file](https://github.com/komish/gh-actions-hackery/blob/c25f0833d32d39ae51efad60e8a3fce5602ab6d4/.github/workflows/skip-behavior.yml).  Job "four" only needs "three", which ran, but "four" will not run because "two" skipped.

As I understand it, the default condition for jobs in workflows is `if: success()`. If I add conditions, they will be appended to the default. E.g.:

```yaml
if: |
  success() &&
  'foo' == 'foo' &&
  ( 'this' == 'this' && 'that' == 'that')
```

If I, however, include my own built-in status check from the array of options (e.g. cancelled(), failure()), then it replaces success(). ([ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions)).

So, including `!cancelled()` here just allows me to run these in the event the skippable job in their execution path skips. The other conditions make sure that the relevant jobs in scope have valid outcomes expected for this job to run. They may be more informational than practical, to your point - because should either of those fail, this job will not run.

---

[1] This is a carryover from the previous implementation, which allowed for empty values for certain environment variables. Given the size of this PR already, I opted to maintain the same design instead of refactoring.

komish:
@mgoerens this PR has a conflict because I've merged another PR, but I'm leaving the conflict unaddressed to give you time to review responses to comments. I'll rebase once you've given me the go ahead.

.github/workflows/release.yml
mgoerens: Thanks for the clarification !

.github/workflows/test.yml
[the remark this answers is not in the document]
mgoerens: Thanks.

I'm happy either way, let's wait for a potential bigger refactor of that logic then.

mgoerens (approved):
/lgtm

jsm84 (approved):
/lgtm

PR #404, opened 2024-11-26
ignore not found errors for ignored files that are not found in dest repo

komish:
If a maintainer removed a file from the repository that was "ignored" in the release_info.json, which controls the syncing of repositories, the release process would fail. In the case a FileNotFoundError is thrown by either the `shutil.rmtree` or `os.remove` commands during the ignore process, we should treat it as non-fatal given the outcome is the same (the file does not exist in the destination repository).


mgoerens (approved):
/lgtm

PR #405, opened 2024-11-26
Consolidate setup-python version definition to one place

komish:
This PR just consolidates the installation of python into a single place in-repo that can be reused across workflows. In practice, this should allow us to update a single place as the python version advances and have it impact all workflows of interest. 

In the future, we can potentially do this exact same thing with the installation of the CI scripts. Task for another day.


mgoerens (approved):
/lgtm

PR #415, opened 2024-12-16
Instanciate new empty lists for each instance of WorkflowRepoManager

### LINKED ISSUE #414: Not found warning when deleting local branches and worktree
The `WorkflowRepoManager` is attempting to deleting some local artifacts that it has created but cannot found them.

Example run: https://github.com/openshift-helm-charts/development/actions/runs/12353276083/job/34472112469?pr=413

```
WARNING:root:local branch "e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45" could not be deleted, potentially because it did not exist. Error: Cmd('git') failed due to: exit code(1)
  cmdline: git branch -D e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45
  stderr: 'error: branch 'e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45' not found'
WARNING:root:local branch "e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat" could not be deleted, potentially because it did not exist. Error: Cmd('git') failed due to: exit code(1)
  cmdline: git branch -D e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat
  stderr: 'error: branch 'e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat' not found'
WARNING:root:local branch "e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat-pr-branch" could not be deleted, potentially because it did not exist. Error: Cmd('git') failed due to: exit code(1)
  cmdline: git branch -D e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat-pr-branch
  stderr: 'error: branch 'e2e-owners-3654720-df65b439e1dc412d9b2e640cdfda0b45-an-owners-file-is-submitted-by-redhat-pr-branch' not found'
WARNING:root:local worktree "<TemporaryDirectory '/tmp/worktree-9w5uv__l'>" could not be deleted, potentially because it did not exist. Error: Cmd('git') failed due to: exit code(128)
  cmdline: git worktree remove /tmp/worktree-9w5uv__l
  stderr: 'fatal: '/tmp/worktree-9w5uv__l' is not a working tree'
  ```

### PR DESCRIPTION
**Pull request by mgoerens**
Previously we were assigning values to attributes on the class level. For list in particular, this had the side effect of leaking values between instances:

```
>>> m = OwnersFileSubmissionsE2ETest()
>>> m.repo_manager._WorkflowRepoManager__local_branches_created
['e2e-owners-d376dfa-6d3820eb5c624ddc9066f96f1a69640e']
>>> m.cleanup()
>>> m.repo_manager._WorkflowRepoManager__local_branches_created
[]
>>> m = OwnersFileSubmissionsE2ETest()
>>> m.repo_manager._WorkflowRepoManager__local_branches_created
['e2e-owners-d376dfa-6d3820eb5c624ddc9066f96f1a69640e', 'e2e-owners-d376dfa-cc7f8acf1850498cb6c758bc286f2bec']
```

Using a dataclass for WorkflowRepoManager, we can ensure that empty lists are initialized for each new instance, and attributes values remain independant.

Fix #414

### DIFF
--- a/tests/functional/behave_features/common/utils/workflow_repo_manager.py
+++ b/tests/functional/behave_features/common/utils/workflow_repo_manager.py
@@ -7,36 +7,27 @@ import git
 
 from tempfile import TemporaryDirectory
 
-import common.utils.github as github
-
-
 class RepoManagementError(Exception):
     pass
 
 
 class WorkflowRepoManager:
-    # Keep a log of things created so we can clean them up.
-    __local_branches_created: [str] = []
-    __local_worktrees_created: [TemporaryDirectory] = []
-    # (remote, branch), e.g. ('openshift-helm-charts/charts, 'my-pr-branch')
-    __remote_branches_created: [(str, str)] = []
-
-    # The token to use for GitHub API operations.
-    __authtoken: str = ""
-
-    # The branch at repo initialization. On Cleanup, we return to this branch
-    # before we remove locally generated branches.
-    original_branch: str = None
-
-    # Working directory at instantiation.
-    old_cwd: str = os.getcwd()
-
-    # The repository at working directory.
-    repo: git.Repo = None
-
     def __init__(self):
         logging.debug(f"{self} --> __init__ called!")
 
+        # Keep a log of things created so we can clean them up.
+        self.__local_branches_created: list[str] = []
+        self.__local_worktrees_created: list[TemporaryDirectory] = []
+        # (remote, branch), e.g. ('openshift-helm-charts/charts, 'my-pr-branch')
+        self.__remote_branches_created: list[tuple(str, str)] = []
+
+        # The token to use for GitHub API operations.
+        self.__authtoken: str = None
+
+        # Working directory at instantiation.
+        self.old_cwd: str = os.getcwd()
+
+        # The repository at working directory.
         try:
             self.repo = git.Repo()
         except git.InvalidGitRepositoryError as e:
@@ -44,6 +35,8 @@ class WorkflowRepoManager:
                 "Unable to initialize git repository. Is the current directory a git repo?"
             ) from e
 
+        # The branch at repo initialization. On Cleanup, we return to this branch
+        # before we remove locally generated branches.
         try:
             self.original_branch = self.repo.active_branch.name
         except TypeError:
@@ -94,6 +87,7 @@ class WorkflowRepoManager:
             self.repo.git.checkout("-b", branch_name)
         except (git.GitCommandError, ValueError) as e:
             raise RepoManagementError("Unable to create branch") from e
+        logging.debug(f'Adding {branch_name} to local_branches_created')
         self.__local_branches_created.append(branch_name)
 
     def add_worktree(self) -> TemporaryDirectory:

### CODE REVIEW

**Review by komish — COMMENTED** [review]
As you know, I'm not a particularly huge fan of dataclasses in class definitions that have a substantial amount of logic. With that said, I think this may solve the problem, but I still feel it changes the contract of the class in a way that may not be strictly required.

The issue with **WorkflowRepoManager** causing this bug is that I heavily used class attributes instead of instance attributes. I did some reading, and this is bug seems to be caused strictly by assumptions I made in how attributes should be defined.

I think the following solves this in the exact same way, without the use of fields or dataclasses. We're not using any additional features of fields and dataclasses, so I'd advocate against using them here.

```python
class WorkflowRepoManager:
    def __init__(self):
        logging.debug(f"{self} --> __init__ called!")

        # Keep a log of things created so we can clean them up.
        self.__local_branches_created: list[str] = []
        self.__local_worktrees_created: list[TemporaryDirectory] = []
        # (remote, branch), e.g. ('openshift-helm-charts/charts, 'my-pr-branch')
        self.__remote_branches_created: list[tuple[str, str]] = []

        # The token to use for GitHub API operations.
        self.__authtoken: str = ""

        # Working directory at instantiation.
        self.old_cwd: str = os.getcwd()

        # The repository at working directory.
        try:
            self.repo = git.Repo()
        except git.InvalidGitRepositoryError as e:
            raise RepoManagementError(
                "Unable to initialize git repository. Is the current directory a git repo?"
            ) from e

        # The branch at repo initialization. On Cleanup, we return to this branch
        # before we remove locally generated branches.
        try:
            self.original_branch = self.repo.active_branch.name
        except TypeError:
            self.original_branch = self.repo.git.rev_parse("--short", "HEAD")
...
```
In other words, if I had just instantiated all of the variables in `__init__(...)`, we probably wouldn't have this buggy behavior. My apologies.

Let me know what you think.

tests/functional/behave_features/common/utils/workflow_repo_manager.py
```diff
@@ -14,15 +15,16 @@ class RepoManagementError(Exception):
...
+@dataclass
 class WorkflowRepoManager:
     # Keep a log of things created so we can clean them up.
-    __local_branches_created: [str] = []
-    __local_worktrees_created: [TemporaryDirectory] = []
+    __local_branches_created: list[str] = field(default_factory=list)
+    __local_worktrees_created: list[TemporaryDirectory] = field(default_factory=list)
     # (remote, branch), e.g. ('openshift-helm-charts/charts, 'my-pr-branch')
-    __remote_branches_created: [(str, str)] = []
+    __remote_branches_created: list[(str, str)] = field(default_factory=list)
```
komish, line 24:
```suggestion
    __remote_branches_created: list[tuple[str, str]] = field(default_factory=list)
```

mgoerens:
Thanks for the review. I'm interested in having a broader discussion around dataclasses, as I'd like to better understand your point of view and wonder if I should use them more sporadically.

> As you know, I'm not a particularly huge fan of dataclasses in class definitions that have a substantial amount of logic. With that said, I think this may solve the problem, but I still feel it changes the contract of the class in a way that may not be strictly required.

WDYM with "it changes the contract of the class" ?

True, the current implementation does change the signature of the (now auto-generated) `__init__` method, by adding the ability to specify values for each attribute. It's possible to do for instance: `w = WorkflowRepoManager(original_branch = "mybranch")`

We didn't need to do that until now, and we most likely won't in the future, so if we want to disable this behavior we can use `init=False` on each [field](https://docs.python.org/3/library/dataclasses.html#dataclasses.field). I'll concede that this is somewhat cumbersome.

> The issue with WorkflowRepoManager causing this bug is that I heavily used class attributes instead of instance attributes. I did some reading, and this is bug seems to be caused strictly by assumptions I made in how attributes should be defined.

Agreed, that matches also with my findings.

> I think the following solves this in the exact same way, without the use of fields or dataclasses [...] ... , if I had just instantiated all of the variables in __init__(...) ...

Yes, I agree that this is the solution (instantiating in `__init__`). That's essentially what is achieved by using a dataclass. This is really the only reason why I went with a dataclass here.

Now I'm wondering if my use of dataclass make the code more or less readable.

Curious about your thoughts.

komish (approved):
/lgtm

PR #418, opened 2025-01-14
Refactor get verify params

mgoerens:
This commit doesn't change the behavior of the get-verify-param step. It
builds upon the new Submission data structure to avoid additional calls
to the GitHub API and improve some readability aspects of the code.

This commit also introduces a slight modification to the Submission data
structure. Previously, the path related to the chart's source pointed to
the "./src/Chart.yaml" file, it now points to the "./src" directory.
This is useful as this is one of the return value of
generate_verify_options.

Finally, this commit also changes the SUBMISSION_PATH environment
variable to an absolute path. This way, steps that run from within the
"./pr-branch" directory such as get-verify-param can easily load the
submission artifact. Note that $GITHUB_WORKSPACE is the default path
under which load an artifact is downloaded by the download-artifact
action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,10 @@ on:
 env:
   # Path to the submission artifact.
   # This artifact is created in the 'validate-submission' job. It contains a JSON representation of
-  # the Submision object. It can be read by other jobs to access those information rather than
+  # the Submission object. It can be read by other jobs to access those information rather than
   # fetch them from the GitHub API a second time.
-  SUBMISSION_PATH: "submission.json"
+  SUBMISSION_PATH: "${{ github.workspace }}/submission.json"
+  SUBMISSION_ARTIFACT_NAME: submission
 
 jobs:
   setup:
@@ -159,7 +160,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ always() && steps.validate-submission.outputs.submission_file_present == 'true' }}
         with:
-          name: submission
+          name: ${{ env.SUBMISSION_ARTIFACT_NAME }}
           path: ${{ env.SUBMISSION_PATH }}
 
       - name: Add 'content-ok' label
@@ -233,7 +234,7 @@ jobs:
       - name: Download submission information
         uses: actions/download-artifact@v4
         with:
-          name: submission
+          name: ${{ env.SUBMISSION_ARTIFACT_NAME }}
 
       - name: Remove 'authorized-request' label from PR
         uses: actions/github-script@v7
@@ -264,7 +265,7 @@ jobs:
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
         working-directory: ./pr-branch
         run: |
-          ../ve1/bin/get-verify-params --directory=pr --api-url=${{ github.event.pull_request._links.self.href }}
+          ../ve1/bin/get-verify-params --directory=pr
 
       - name: Install oc
         id: install-oc
@@ -429,7 +430,7 @@ jobs:
       - name: Download submission information
         uses: actions/download-artifact@v4
         with:
-          name: submission
+          name: ${{ env.SUBMISSION_ARTIFACT_NAME }}
 
       # Chart report errors are only present if an error has occured during the chart_report
       - name: Download chart_report errors
--- a/scripts/src/report/get_verify_params.py
+++ b/scripts/src/report/get_verify_params.py
@@ -2,84 +2,64 @@ import argparse
 import os
 import sys
 
+from submission import submission, validate
+
 sys.path.append("../")
 from chartprreview import chartprreview
 from signedchart import signedchart
 from tools import gitutils
 
 
-def get_report_full_path(category, organization, chart, version):
-    return os.path.join(
-        os.getcwd(), get_report_relative_path(category, organization, chart, version)
-    )
+def generate_verify_options(directory: str, s: submission.Submission):
+    """Prepare a set of GitHub outputs, including the options for chart-verifier verify
 
+    Args:
+        directory (str): Local directory in which to write the error logs
+        s (Submission): Submitted GitHub Pull Request
 
-def get_report_relative_path(category, organization, chart, version):
-    return os.path.join("charts", category, organization, chart, version, "report.yaml")
+    Returns:
+        str: The flags used by chert-verifier verify command
+        str: The path to the chart (tarball or source)
+        bool: Whether a report needs to be generated
+        bool: Whether a cluster is needed
 
+    """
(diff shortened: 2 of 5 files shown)


scripts/src/report/get_verify_params.py
```diff
@@ -2,84 +2,64 @@
...
         cluster_needed = False
-        report_provided = True
 
-    if os.path.exists(src) and not os.path.exists(tar):
+    if s.source.found and not s.tarball.found:
         print("[INFO] chart src included")
-        return flags, src, True, cluster_needed, report_provided
-    elif os.path.exists(tar) and not os.path.exists(src):
+        src_full_path = os.path.join(os.getcwd(), s.source.path)
+        return flags, src_full_path, True, cluster_needed
```
komish, line 39: In your incoming return statement, I see the returned parameters have changed order and meaning. Yours:

```python
return flags, src_full_path, True, cluster_needed
```

previous:

```python
return flags, src, True, cluster_needed, report_provided
````

Specifically **cluster_needed** and **report_provided** seem to be inverted. In addition, the previous boolean state of the report seems inverted as well. That is **report_provided** (previous) is the inverse of **report_needed** (yours).

I'd say that these don't technically need to change for a preexisting API, and that changing them has a higher degree of likelihood of causing inadvertent side effects. Have you reviewed all callers of this function to make sure their use of returned values aligns with the new definition? Alternatively, could we implement this refactor without making these contractual changes?

File: scripts/src/submission/submission.py [hunk 2]
```diff
@@ -188,7 +188,7 @@ class Report:
 @dataclass
 class Source:
     found: bool = False
-    path: str = None  # Path to the Chart.yaml
+    path: str = None
```
komish, line 191: I'd like to see an example of what this path is expected to look like in a comment.

You haven't touched other code in this PR, so I'll just mention it as an improvement opportunity here.

If it's an absolute path, calling the attribute "full_path" may not be out of scope either, but it isn't required to accomplish what I'd look for in the comment.

scripts/src/submission/submission_test.py
```diff
@@ -305,7 +305,7 @@ class CertificationScenario:
         expected_is_valid_certification=False,
         expected_reason="[ERROR] Send OWNERS file by itself in a separate PR.",
     ),
-    # Invalid certification Submission contains OWNERS and report file, but ignore_owners is set to True
+    # Valid certification Submission contains OWNERS and report file, but ignore_owners is set to True
```
komish, line 308: Any reason why the comment's description of the test is inverting, but no code is changing? Just an error in the comment, but the test actually tests a valid condition?

scripts/src/report/get_verify_params.py
```diff
@@ -2,84 +2,64 @@
...
-
-
-def get_report_relative_path(category, organization, chart, version):
-    return os.path.join("charts", category, organization, chart, version, "report.yaml")
+def generate_verify_options(directory: str, s: submission.Submission):
+    """Prepare a set of GitHub outputs, including the options for chart-verifier verify
 
+    Args:
+        directory (str): Local directory in which to write the error logs
+        s (Submission): Submitted Github Pull Request
```
komish, line 18: NIT, feel free to ignore.

```suggestion
        s (Submission): Submitted GitHub Pull Request
```

.github/workflows/build.yml
```diff
@@ -159,7 +160,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ always() && steps.validate-submission.outputs.submission_file_present == 'true' }}
         with:
-          name: submission
+          name: ${{ env.SUBMISSION_ARTIFACT_NAME}}
```
komish, line 163:
```suggestion
          name: ${{ env.SUBMISSION_ARTIFACT_NAME }}
```

.github/workflows/build.yml
```diff
@@ -9,7 +9,8 @@ env:
   # This artifact is created in the 'validate-submission' job. It contains a JSON representation of
   # the Submision object. It can be read by other jobs to access those information rather than
   # fetch them from the GitHub API a second time.
-  SUBMISSION_PATH: "submission.json"
+  SUBMISSION_PATH: "${{ github.workspace }}/submission.json"
+  SUBMISSION_ARTIFACT_NAME: submission
```
komish, line 13: 1.NIT: In the comment, `Submission` is misspelled. 
2. should SUBMISSION_ARTIFACT_NAME have a file extension? It's still a JSON file, no?
3. Should SUBMISSION_PATH use SUBMISSION_ARTIFACT_NAME in its definition?

.github/workflows/build.yml
```diff
@@ -264,7 +265,7 @@ jobs:
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
         working-directory: ./pr-branch
         run: |
-          ../ve1/bin/get-verify-params --directory=pr --api-url=${{ github.event.pull_request._links.self.href }}
+          ../ve1/bin/get-verify-params --directory=pr
```
komish, line 268: Why is this changing?

komish (changes requested):
I'd like to know more about what's driving this change? The key thing I'm seeing is that operations are now relying on a populated Submission object, which itself is expected to contain **all** metadata related to the pull request. And so, we delegate the responsibility of figuring that out to the Submission artifact instead of implementing logic in get-verify-params which makes calls to populate the same information. Is that the correct read?

I also think there's an opportunity to implement some sort of "verifier flag builder" class, but I think that's a story for another day.

scripts/src/submission/submission.py
mgoerens: I've added some docstrings and comments

scripts/src/report/get_verify_params.py
mgoerens: Done

mgoerens (commented):
> I'd like to know more about what's driving this change? The key thing I'm seeing is that operations are now relying on a populated Submission object, which itself is expected to contain all metadata related to the pull request. And so, we delegate the responsibility of figuring that out to the Submission artifact instead of implementing logic in get-verify-params which makes calls to populate the same information. Is that the correct read?

That's essentially correct.

The submission artifact is populated once at the beginning of the pipeline with all information that are needed by later steps.

For example, there was some logic to determine the category, name, ... of the chart, whether a tarball or a source has been provided, etc. This kind of information gathering and validation don't have to be repeated all over the place. We can now rely on the Submission artifact with information that have been validated already and that we can trust.

> I also think there's an opportunity to implement some sort of "verifier flag builder" class, but I think that's a story for another day.

Agreed. I didn't find a clever way to do that in this commit, and didn't want to overload the PR . Another day :)

komish (approved):
/lgtm

Issue #432, opened 2025-07-01
CI pipeline fail due to GITHUB_REF variable not being populated


mgoerens:
This occurs regularly in the nighlty runs. A simple re-run usually fixes it.

Example run: https://github.com/openshift-helm-charts/sandbox/actions/runs/15991583337/job/45105893057?pr=127783

What seems to happen is that for some reason the `GITHUB_REF` variable is not correctly populated by GitHub.

```
INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "gh-pages"; else echo "${GITHUB_REF##*/}-gh-pages"; fi)
```

As a result, the `INDEX_BRANCH` becomes `-gh-pages` and `git fetch upstream "$INDEX_BRANCH"` fails miserably with the error message `error: unknown switch 'g'`

mgoerens:
Following #433, we can confirm that the `GITHUB_REF` (and `github.ref`) variable being empty is the direct cause of the failure: [example failed run with debug](https://github.com/openshift-helm-charts/sandbox/actions/runs/16084338261/job/45393584102?pr=128173)

Now, why is GITHUB_REF empty ? From (the github doc)[https://docs.github.com/en/actions/reference/contexts-reference#github-context]:

> github.ref: [...] This is only set if a branch or tag is available for the event type. [...]

Under certain conditions, it is expected to be empty. However, those conditions are not met here AFAICT. Also, on a passing run with similar conditions (or even on a re-try) it is correctly populated.

This indicates a bug with the GitHub platform, not our workflow.

@komish It is difficult to debug further here. I also doubt that bringing this up to GitHub is going to provide a resolution, let alone a quick one. If I had to guess, I would say that the number of objects in the sandbox repo (over 100000 PRs) is a possible cause of this bug. We already know that browsing the repo via WebUI can be very slow. It could very well be causing internal timeouts during the workflow initialization causing the conditions for populating `github.ref` not to be met. I could try to collect additional data to support this theory (e.g. analyze workflow initialization) but this doesn't feel like a really good use of my time, as we won't be able to validate this theory 100% anyway.

Note also that we never experienced this issue in the prod repo, only in the sandbox repo.

WDYT ? Do you have an idea, a theory ?

FYI, I already tried in the past to ask GitHub to permanently delete the older PRs. Unfortunately, they only carry this action if there is a security or confidentiality risk. One solution would be to recreate the sandbox repo from scratch.

[...]

PR #433, opened 2025-07-01
Add debug for tracking GITHUB_REF value

mgoerens:
Debug to better understand #432


mgoerens:
@komish More debug ! Yes ! Why not, with those kind of weird issues, you never know

PR #448, opened 2025-08-20
use new RateLimitOverview class to check GitHub rate limit information

komish:
The PyGithub library was updated in this pull request: https://github.com/openshift-helm-charts/development/pull/439.,

It introduced a documented breaking change when requesting rate limit information via the Github object. This pull request implements the changes suggested in their [changelog](https://github.com/PyGithub/PyGithub/blob/ad278c5f33931c33f374de5d03a3f3d0831c4aaf/doc/changes.rst?plain=1#L13-L25) and adds some documentation while I'm working with this function.

Should resolve issues with the "Scheduled Metrics" workflow.

### DIFF
--- a/scripts/src/metrics/metrics.py
+++ b/scripts/src/metrics/metrics.py
@@ -549,10 +549,11 @@ def send_metric(write_key, id, event, properties):
     analytics.track(id, event, properties)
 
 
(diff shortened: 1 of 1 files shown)


jsm84 (approved):
lgtm

PR #450, opened 2025-08-25
Ensure submission category metadata is in place for OWNERS-only submissions

komish:
OWNERS file modifications made by partners would receive the typical response for non-partners (e.g. "must be reviewed by maintainers") instead of instructing them to make changes in their partner dashboard. This seemed to be caused by the submission category not being populated because of this missing matcher, which reads the submission path. Should be resolved here.


mgoerens (approved):
Good catch. LGTM

PR #460, opened 2025-10-16
Get metadata from submission object

mgoerens:
This is a follow-up of #456:
- Remove unused code
- Move the output of category/organization/name to the check_for_owners task, as we have all the information we need from the submission object


komish (approved):
/lgtm

When this releases, please do a double check on mercury-bot partner OWNERS file updates to make sure that they continue to behave as expected. It's a workflow that's relatively difficult to validate in our tests because of the required user.

PR #479, opened 2025-11-06
Fix error message for PRs with OWNERS and chart

### LINKED ISSUE #477: Incorrect error message when user submits both a chart and an OWNERS
Example run: https://github.com/openshift-helm-charts/charts/actions/runs/19066229904/job/54457476435?pr=2487

This PR contains an OWNERS file + a chart, the error states:

> [A PR must contain only one chart. Current PR includes files for multiple charts.](https://github.com/openshift-helm-charts/development/blob/main/scripts/src/submission/submission.py#L106)

Which is not exactly on point. It should be:

> [Send OWNERS file by itself in a separate PR.](https://github.com/openshift-helm-charts/development/blob/main/scripts/src/submission/submission.py#L431)

The problem seems to come from [this condition](https://github.com/openshift-helm-charts/development/blob/main/scripts/src/submission/submission.py#L100-L104). When we register an OWNERS file, it doesn't relate to any version, hence it is flagged as a different chart.

mgoerens:
PRs containing a chart and an OWNERS file were getting the wrong error message. This was due to the check for duplicate charts, not taking into account that an OWNERS file doens't relate to a specific version, and thus considering the OWNERS file to be for a different chart.

Specific tests have been added to cover this case.

Fix #477

### DIFF
--- a/scripts/src/submission/submission.py
+++ b/scripts/src/submission/submission.py
@@ -101,7 +101,7 @@ class Chart:
             (self.category and self.category != category)
             or (self.organization and self.organization != organization)
             or (self.name and self.name != name)
-            or (self.version and self.version != version)
+            or (self.version and version and self.version != version)
         ):
             msg = "[ERROR] A PR must contain only one chart. Current PR includes files for multiple charts."
             raise DuplicateChartError(msg)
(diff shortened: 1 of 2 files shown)


komish:
This is relatively small but I'll wait until next release to merge it to keep the fast-following release (pushing out token permission updates) tiny.

PR #500, opened 2026-03-12
Enable smoke testing on PR label

komish:
Repo maintainers can now label a pull request with the trusted label to allow the smoke testing of the content of a pull request. Original intent is to allow a maintainer to run tests against a dependabot PR.

NOTE:

When dependabot creates a pull requests, there are two test-related workflows that start: release.yml and test.yml.

release.yml fails because it's configured to fail outright if the contributor is not an approved contributor. This will continue to be the case for now when dependabot submits a PR, and will need to be reworked in the future.

test.yml succeeds without running any tests (which effectively means it just skips testing). Therefore, this is not currently seen as a failing test when dependabot submits a pull request.

I think there's an opportunity to refactor both release.yml and test.yml to utilize the label-based approach I'm implementing here, but until then, we still want to be testing dependabot PRs so that we do not have to recreate them for them to be tested as seen in #497.

### DIFF
--- a/.github/workflows/test-smoke-on-label.yml
+++ b/.github/workflows/test-smoke-on-label.yml
@@ -1,0 +1,104 @@
+name: Smoke Test
+
+# Runs smoke testing on pull requests that have the trusted label. Useful for
+# cases where the automated pull request testing won't run because the
+# contributor is not approved. E.g. Dependabot PRs.
+#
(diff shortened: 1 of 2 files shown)


jsm84 (approved):
Reviewed and LGTM.

PR #511, opened 2026-04-30
Add github-actions ecosystem to dependabot

komish:


mgoerens (approved):
LGTM

I notice you did use `groups` this time. Fine by me, as the new action version is actually tested (vs chart-verifier)

komish:
@mgoerens Yep, made sense here just for that reason. At some point, if we move certification-specific actions into a subdirectory, I'll also plan on creating a second group just for those to separate the dependabot work.

Unrelated: I'll wait for release completion to merge this.

PR #512, opened 2026-04-30
Disable stage and prod release PR automerging

komish:
We run tests in stage and prod when releases are cut, but would rather have a human control the merge so that we can tier merges to stage and prod.

Development release PRs are merged by release.yml, removing the merge step from here should only prevent auto merging in the stated repositories. We leave the approval step for clarity.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,8 +139,8 @@ jobs:
       bot-name: ${{ secrets.BOT_NAME }}
       bot-token: ${{ secrets.BOT_TOKEN }}
 
-  approve-and-merge:
-    name: Approve and merge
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
caxu-rh: Nit: might want to consider adjusting job slug and human-readable `name` to reflect that this job is now only approving, not merging.

komish: Oversight on my part. Good catch. https://github.com/openshift-helm-charts/development/commit/316e618a1220431c57a1f06c8f71d14cc6c5cc43 should have that corrected.

caxu-rh (approved):

PR #522, opened 2026-06-02
Specify valid conditions for authorized-request label application

komish:
The authorized-request label would be applied to PRs in cases where the chart-verifier job would not complete due to some transient error. Actual user verification is delegated to a script, but we can also prevent this label from being applied by being more specific in what job outcomes matter to us (done here by ensuring that the outcome is either success or failure, and not allowing an empty value.


mgoerens:
If I understand correctly, `needs.chart-verifier.outputs.run-verifier-outcome` could be empty thus passing the `!= failure` test.

/lgtm

mgoerens (approved):

PR #524, opened 2026-07-14
Address CI GPG chart signing key expiration

komish:
Generates a new GPG key that is not expired for use in tests, now that the Helm verification library throws an error for expired keys.

Test fixture has been resigned, generating the provenance file included. Should resolve CI failures for case HC-10.

EDIT: @jsm84 / maintainers - The reason [HC-10](https://github.com/openshift-helm-charts/sandbox-2025-11/pull/14098) _continues to fail in this PR_ is because the test case points to the vault test fixture that's included in [this repository via the default branch](https://github.com/openshift-helm-charts/development/blob/main/tests/data/HC-10/signed_chart/report/partner/report.yaml#L10). Until this PR merges, that test fixture in the main branch will show that it's signed by the old key. Once this merges, the HTTP endpoint should then return the vault chart signed with the new key, allowing the test case to pass.


jsm84 (approved):
LGTM

